PR #6, opened 2022-03-15
chore: 👷 Add Husky

melMass:
Each commits get autoformatted with `prettier`,
could be used for other things.

https://typicode.github.io/husky


melMass:
I sneaked 6d2b3daf644738d4ef9c5f2dd38c46840eb0220b in, it's not really related but it was to avoid another PR

.husky/pre-commit
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+echo "Formating project and committing..."
+FILES=$(git diff --cached --name-only --diff-filter=ACMR | sed 's| |\\ |g')
+[ -z "$FILES" ] && exit 0
+
+# Prettify all selected files
+echo "$FILES" | xargs ./node_modules/.bin/prettier --ignore-unknown --ignore-path .gitignore --write --plugin-search-dir=.
```
xat: @melMass do you think we should limit this to .js, .jsx and maybe .scss files?

melMass: It should be handled by prettier but we can restrict it before if needed

xat:
@melMass i pulled in pretty-quick, this made it a bit easier to use the same file pattern in the hook as in the npm script.

i also changed the npm script from `npm run prettier` to `npm run format`, think that's a more convenient name and added `.scss` to the list of supported files.

after adding `.scss`, i formatted all scss files.

melMass:
Awesome! Very nice! 
I think we can merge?

xat (approved):
great idea :)

PR #20, opened 2022-03-18
enabled teia contract

xat:
Bildschirmfoto 2022-03-18 um 16 22 49


melMass:
Looks good!
Do you have tokens to test it?

Here is one I can reswap to test out:

https://teia-ui-main-pr-20.onrender.com/objkt/674903

Edit: I'm dumb we can test on any objkt, this one was just already swapped on teia in the past and is a good test case as is has 0% royalties.

So yes I tested on other objkts and I have the same issue 🤷‍♂️ !

melMass:
I just tested on the link and locally, and when I set my swap options (1ed, 500tez) it displays `Preparing Swap` for a second and then goes back to the objkt display, without calling the wallet

xat:
> I just tested on the link and locally, and when I set my swap options (1ed, 500tez) it displays `Preparing Swap` for a second and then goes back to the objkt display, without calling the wallet

yeah, that was a stupid mistake on my end. fixed now :)

i also changed the indexer to the main indexer.

melMass:
@jagracar Why is there two calls to `update operators` ?
image

@xat Nice! Now you need to buy it.... just for the test 🤣
image

melMass:
Btw! imo we can't ship with this reswap field as is if we announce an "UI update".
I will file an issue and try to tackle it today!

melMass (approved):
As far as I can test it it all seems to work fine! 
And with the prod indexer 😍

scalemodal:
> 
@melMass 
Could we apply design updated styles for this feature as well?
Input field and buttons like this:

image

Styles: https://www.figma.com/file/R5sw1HcoHx4Aup1Yl7rGYA/UX%2FUI?node-id=1406%3A11773

melMass:
> @melMass Could we apply design updated styles for this feature as well? 

Let's use #22 to file the issues, it might require to be split in a few PRs
It's not related to teia's contract and this is already used in prod... Just very little devs were aware of this feature/implementation afaik

PR #23, opened 2022-03-19
Make collab names/addresses clickable

### LINKED ISSUE #19: [feat] make collab username clickable
on the signing status modal, it would be great if the usernames were clickable to go to that profile directly... would likely reduce confusion and risk of scams

(submittet by aebrer on discourse)

collabclick

### LINKED ISSUE #31: [feat] signing summary missing, since update
looks like the inline dialog that shows a list of signers got lost due to the update. we should re-add it.

this here is currently missing:
Bildschirmfoto 2022-03-20 um 13 28 13

melMass:
* fix: ✨ fixes #19, prevent event bubbling

This is half a solution to the problem of preventing event bubbling.

### DIFF
--- a/src/components/collab/show/CollabIssuerInfo.js
+++ b/src/components/collab/show/CollabIssuerInfo.js
@@ -20,16 +20,14 @@ export const CollabIssuerInfo = ({ creator }) => {
       <a
         className={styles.issuerBtn}
         href={path}
-        onMouseEnter={() => setShowCollabSummary(true)}
+        onMouseOver={() => setShowCollabSummary(true)}
+        onMouseOut={() => setShowCollabSummary(false)}
       >
         <Primary>{name !== '' ? name : walletPreview(address)}</Primary>
       </a>
 
       {showCollabSummary && (
-        <div
-          className={styles.collabInfo}
-          onMouseLeave={() => setShowCollabSummary(false)}
-        >
+        <div className={styles.collabInfo}>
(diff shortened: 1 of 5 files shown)


aebrer (approved):
looks great to me! 😄

aebrer:
only potential issue is the verification checkmarks don't seem to be showing, but that's true on the prod site anyway, so I'm not sure it's related to this directly

melMass:
> only potential issue is the verification checkmarks don't seem to be showing, but that's true on the prod site anyway, so I'm not sure it's related to this directly

It is indeed coming from before... But it's a good candidate to also fix in this PR.
I will look at xat's fix in roughly an hour and compare it with before we merged the UI WG PRs!

xat:
looks good to me. what do you think @melMass ?

xat:
@aebrer if you think it's fine, can you approve the PR? you are set as reviewer who needs to approve, before we can merge :)

aebrer (approved):
looks good to me :)

aebrer:
@xat I'm not sure my review counts though, it still says it needs an approved review

xat (approved):

xat:
@aebrer ah true, apparently i also needed to approve, even though i wasn't set as a reviewer :-)

@melMass if there is nothing left from your side, i think it can be merged.

melMass:
> @melMass if there is nothing left from your side, i think it can be merged.

Good to go for now for me!
It will need refactoring once we are there

PR #25, opened 2022-03-19
fix: ⚡️ Fix dependency vulnerabilities

### LINKED ISSUE #26: [bug] "random" button doesn't respond
the "random" button on the teia main feed prints out a blank feed.
(since the UI update)

melMass:
## Disclaimer

This is updating a lot of dependencies. To properly test it locally if you have already pulled the repo before, you must first delete `node_modules` and re-run `npm i`

---
Current `main`:
```ruby
50 vulnerabilities (27 moderate, 20 high, 3 critical)
```

This PR:
```ruby
10 vulnerabilities (7 moderate, 3 high)
```

---
I'm following dependabot alerts for now.
Most of them actually came from the templates.

Fixes:
[dependabot/2](https://github.com/teia-community/teia-ui/security/dependabot/2)
[dependabot/9](https://github.com/teia-community/teia-ui/security/dependabot/9)
[dependabot/50](https://github.com/teia-community/teia-ui/security/dependabot/50)
[dependabot/26](https://github.com/teia-community/teia-ui/security/dependabot/26)


.env
```diff
@@ -1,3 +1,5 @@
+GENERATE_SOURCEMAP=false
```
melMass, line 1: This is due to that: 
https://github.com/facebook/create-react-app/discussions/11767

src/utils/verify.js
```diff
@@ -0,0 +1,27 @@
+var sodium = require('sodium-universal')
```
melMass: I extracted the functions we make us of only from the ez lib thing which had many security issues associated. 

This seems to still error out. I need to investigate further

melMass:
(it's not ready for review, I'm trying to trigger the action)

melMass:
Ready to review.
IMO we should merge it and not Squash it because there is a lot going on, I tried to separate the commits properly

src/utils/verify.js
melMass: Solved in 2ec678975c6f9ca07d77dc4ef4e72db6b8725462

melMass:
@xat Do you know why there is no action running anymore?

xat:
hmm, the action did work in the pr that i just opened. maybe a temp problem?

xat:
@melMass i would suggest that we wait a week or so, before merging this one after we ironed out the more pressing bugs that got revealed due to the bigger release. because updating dependencies can bring in further instability, which i would want to avoid until things are bit more settled.

would it make sense to move the "get random objkts" commit into a separate pr, then we could merge that one upfront?

what do you think?

melMass:
> @melMass i would suggest that we wait a week or so, before merging this one after we ironed out the more pressing bugs that got revealed due to the bigger release.


Yes it makes sense, I spent hours on it so I won't have the same time to dedicate fixing it in a week, but we could use it as a reference or something like that 👌

xat:
@melMass ah ok, so delaying the merge would create extra effort.

hmm, do you think it is safe enough to merge it sooner? my feeling is that it does require some extended testing.

melMass:
> @melMass ah ok, so delaying the merge would create extra effort.
> 
> 
> 
> hmm, do you think it is safe enough to merge it sooner? my feeling is that it does require some extended testing.

Yes it definitely need more extensive testing, but I meant more if we fix other stuffs under it it might be better to start this one over, following it as a ref. I spent most of my time checking breaking changes in libraries, I know them now

melMass:
Rebased 😏  🤣

I'm going to try to maintain it as a virtual `develop` branch and rebase regularly until someone can look into it/ we can test it further

craco.config.js
```diff
@@ -0,0 +1,33 @@
+const webpack = require("webpack");
+const path = require("path");
+module.exports = {
+  webpack: {
+    alias: {
```
xat, line 5: nice 👍

package.json
```diff
@@ -3,49 +3,48 @@
...
-    "@taquito/michelson-encoder": "^9.1.1",
-    "@taquito/taquito": "^10.1.3",
-    "@testing-library/jest-dom": "^5.14.1",
-    "@testing-library/react": "^11.2.7",
-    "@testing-library/user-event": "^13.1.9",
-    "axios": "^0.21.1",
+    "@airgap/beacon-sdk": "^2.3.10",
+    "@taquito/beacon-wallet": "^12.0.0",
+    "@taquito/michelson-encoder": "^12.0.0",
+    "@taquito/taquito": "^12.0.0",
```
xat, line 9: did you check if there were any breaking changes?

package.json
```diff
@@ -3,49 +3,48 @@
...
+    "libsodium-wrappers": "^0.7.9",
     "local-storage": "^2.0.0",
     "lodash": "^4.17.21",
-    "markdown-to-jsx": "^7.1.3",
-    "mime-types": "^2.1.30",
-    "node-fetch": "^2.6.1",
-    "node-sass": "^6.0.1",
+    "markdown-to-jsx": "^7.1.7",
+    "mime-types": "^2.1.35",
+    "nanoassert": "^2.0.0",
```
xat: is this used?

package.json
```diff
@@ -66,8 +65,17 @@
     ]
   },
   "devDependencies": {
-    "prettier": "^2.3.0",
-    "husky": "^7.0.0"
+    "@craco/craco": "^7.0.0-alpha.3",
+    "assert": "^2.0.0",
```
xat, line 68: why were these libraries not needed before?

src/components/collab/create/CollaboratorRow.js
```diff
@@ -1,11 +1,11 @@
 import { useEffect, useRef, useState } from 'react'
-import { Button, Secondary } from '../../../components/button'
-import {} from '../../media-types/'
+import { Button, Secondary } from '@components/button'
+import {} from '@components/media-types/'
```
xat, line 3: import can be removed

src/data/hicdex.js
```diff
@@ -207,3 +209,62 @@ export async function fetchGraphQL(operationsDoc, operationName, variables) {
...
+    }`,
+    'LastId'
+  )
+  return data.hic_et_nunc_token[0].id
+}
+
+export async function fetchRandomObjkts(count) {
+  const firstId = 196
+  const lastId = await getLastObjktId()
+  console.log(`Last id is : ${lastId}`)
```
xat: console.logs in the file probably can be removed

src/data/hicdex.js
```diff
@@ -207,3 +209,62 @@ export async function fetchGraphQL(operationsDoc, operationName, variables) {
...
+        id
+      }
+    }`,
+    'LastId'
+  )
+  return data.hic_et_nunc_token[0].id
+}
+
+export async function fetchRandomObjkts(count) {
+  const firstId = 196
```
xat, line 296: unrelated to this pr, but afaik 152 is the first objkt id

src/data/hicdex.js
```diff
@@ -207,3 +209,62 @@ export async function fetchGraphQL(operationsDoc, operationName, variables) {
...
+  while (uniqueIds.size < count) {
+    const id = rnd(firstId, lastId)
+    console.log(id)
+    uniqueIds.add(id)
+  }
+  try {
+    let objkts = await fetchObjkts(Array.from(uniqueIds))
+    return objkts
+  } catch (e) {
+    console.error(e)
```
xat, line 306: would it make sense to re-throw the error here?

xat:
looks very nice!

melMass:
Ok now it's ready to be reviewed for real @xat 😇

xat:
from the first review, there are few (small) open questions, but otherwise it looks good!

maybe some non-devs can do a bit of manual testing before we merge it.

melMass:
@xat Mmhh so while debugging the feed thing (that's actually search), it seems src/pages/feed isn't used anymore?

xat:
> @xat Mmhh so while debugging the feed thing (that's actually search), it seems src/pages/feed isn't used anymore?

yeah, looks like `/random`, `/latest` and `/hdao` are dead routes

xat (approved):
can be merged imo, if enough testing was done

xat:
nice work :>

package.json
melMass: Yep that's the whole idea of this PR

package.json
melMass: I'm not sure like that I'm going to check in context!

package.json
melMass: [Craco](https://github.com/gsoft-inc/craco) is what allows all I'm doing here: extend CRA without ejecting

melMass: Assert for the polyfills, I'm not 100% sure it's needed but if not used I don't think it adds anything to the bundle

src/data/hicdex.js
melMass: I'm going to do a round of console.log indeed! I've seen others while checking the console.

src/data/hicdex.js
melMass: As far as I can tell this value was always set to that. I wonder what is between 152 and 196 maybe tests?

src/data/hicdex.js
melMass: I think one of the refactor aspects will be to standardize that. Some return undefined, some log errors, some throw!
I agree they should all throw and excpetions should be handled where it's used

src/data/hicdex.js
melMass: Mmm I'll just remove those you're right, it will complicate rebases on top. But I think we can remove 90% of them 😅

PR #30, opened 2022-03-20
fixed swapping in collab mode

xat:
this fixes swapping in collab mode. the operator in the update_operators transaction was wrongly set to the address of the collab contract instead of the marketplace address.

this should carefully be tested, before merging.

### DIFF
--- a/src/context/HicetnuncContext.js
+++ b/src/context/HicetnuncContext.js
@@ -94,7 +94,7 @@ function createSwapCalls(
   objktsContract,
   marketplaceContract,
   objktsAddress,
-  marketplaceAddress,
+  operatorAddress,
   objkt_id,
   ownerAddress,
   objkt_amount,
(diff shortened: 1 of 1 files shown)


xat:
ztepler and i tested the implementation and it worked.

@melMass if you can also do a final test, i think it would be good to merge afterward.

ztepler:
Thank you, just checked it on https://teia-ui-pr-30.onrender.com/, was able to `swap` and `unswap` token, transactions prepared good, works fine. Also code diff looks valid too (thank you for all clarifications on discord)

melMass:
> ztepler and i tested the implementation and it worked.
> 
> 
> 
> @melMass if you can also do a final test, i think it would be good to merge afterward.

I'm afraid I really can't help when it come to the collab contract, I really don't understand much of the code, and even the contract itself.. sorry

xat:
@melMass ok, then maybe someone else can do a final test.

xat:
on the other hand, it was tested by two people... so that should probably be enough.

xat:
MerchantCoppola tested this one as well, so i think it should be good to merge.

melMass (approved):

PR #32, opened 2022-03-22
fixed endless loading of collab tab and redirect to /tz problem

### LINKED ISSUE #2: Collabs page displays a loading indicator forever when you don't have collabs.

xat:
this fixes the problem mentioned in https://github.com/teia-community/teia-ui/issues/2 and a further problem that triggered a redirect to '/tz' when clicking on the collab tab as a user without subjkt.

### DIFF
--- a/src/pages/display/index.js
+++ b/src/pages/display/index.js
@@ -1071,6 +1071,13 @@ export default class Display extends Component {
             onLoaded={this.disableLoading}
           />
         )}
+        {this.state.collabsState && !this.state.loading && (
+          <Container>
+            <Padding>
+              <h1>No collabs</h1>
+            </Padding>
+          </Container>
+        )}
 
         {!this.state.loading && this.state.collectionState && (
           <div>
--- a/src/pages/display/index.js
+++ b/src/pages/display/index.js
@@ -22,6 +22,12 @@ import { getWalletBlockList } from '../../constants'
 const axios = require('axios')
 const fetch = require('node-fetch')
 
+const TAB_CREATIONS = 'creations'
+const TAB_COLLECTION = 'collection'
+const TAB_COLLABS = 'collabs'
+
+const TABS = [TAB_CREATIONS, TAB_COLLECTION, TAB_COLLABS]
+
 const query_collection = `
 query collectorGallery($address: String!) {
   hic_et_nunc_token_holder(where: {holder_id: {_eq: $address}, token: {creator: {address: {_neq: $address}}}, quantity: {_gt: "0"}}, order_by: {token_id: desc}) {
@@ -264,6 +270,16 @@ async function fetchBalance(addr) {
   return result
 }
 
+export function EmptyTab({ children }) {
+  return (
+    <Container>
+      <Padding>
+        <h1>{children}</h1>
+      </Padding>
+    </Container>
+  )
+}
+
 export default class Display extends Component {
   static contextType = HicetnuncContext
 
@@ -283,14 +299,11 @@ export default class Display extends Component {
     objkts: [],
     creations: [],
     collection: [],
-    collabs: [],
     forSale: [],
     notForSale: [],
     marketV1: [],
     items: [],
-    creationsState: true,
-    collectionState: false,
-    collabsState: false,
+    currentTab: TAB_CREATIONS,
     collectionType: 'notForSale',
     showUnverifiedCollabObjkts: false,
     hdao: 0,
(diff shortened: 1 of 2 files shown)


src/pages/display/index.js
```diff
@@ -579,12 +579,11 @@ export default class Display extends Component {
...
     }
 
-    this.updateLocation('collabs')
-  }
-
-  updateLocation = (slug) => {
-    const { subjkt, wallet } = this.state
-    this.props.history.push(`/${subjkt === '' ? wallet : subjkt}/${slug}`)
+    if (this.state.subjkt !== '') {
+      this.props.history.push(`/${this.state.subjkt}/collabs`)
```
melMass: Can't we just always resolve to the address?

xat: hmm, it's done like this in the other tabs as well and ensures that the url contains the username.

melMass: Ah my bad, I thought routing handled these

LeonNicholls (approved):
LGTM

LeonNicholls (approved):

melMass:
@xat Very nice, do you think it's out of scope of this issue to either dim the tab title, or show `No Collaborations` in the "grid view" ?

melMass (commented):
@xat Very nice, do you think it's out of scope of this issue to either dim the tab title (or hide?), or show No Collaborations in the "grid view" ?

melMass:
@xat I made a simple proposal in 6151985157c5c1889037acfe6e7b930f55a25c2a

image

melMass:
Also added collections, discard loading components if the objkts length === 0

image

melMass:
This might more be out of scope for this issue,
but the first render of an artist page is strange, to reproduce, go to the feed: https://teia-ui-pr-32.onrender.com/
Click on an artist name down an objkt, there! something strange happens on "route resolution"
The name is already resolved so it should be used as the path when there is a SUBJKT

melMass:
(rebased)

src/pages/display/index.js
```diff
@@ -1062,151 +1065,166 @@ export default class Display extends Component {
...
+          <CollabsTab
+            wallet={this.state.wallet}
+            filter={this.state.filter}
+            onLoaded={this.disableLoading}
+          />
         )}
-
-        {!this.state.loading && this.state.collectionState && (
-          <div>
+        {((this.state.collabsState && this.state.collabs.length === 0) ||
```
xat: collabs length is always 0 because they are loaded in the CollabsTab component. this means currently always 'No Collabs' is shown.

Bildschirmfoto 2022-03-23 um 08 09 11

xat: fixed

xat:
did some small refactorings on the tab logic, moved the empty-state of the collabs tab into the CollabsTab component, and added the empty-state to the creations tab as well.

this means another round of testing is needed :-)

i think this whole component needs heavy refactoring and would greatly benefit from a library like `useSWR`.

yes, the routing is somehow odd and also needs refactoring. the link to the artist subpage from the start-page should directly point to the slug-version of the artist profile if the artist has a slug, so that these redirects are not needed in the first place. but imo fixing that is out of scope for this pr.

melMass:
> i think this whole component needs heavy refactoring and would greatly benefit from a library like `useSWR`.

Yes my first commit I realised I did not understand it was a huge "if" list 😅 .....
Thanks for the fix! Another thing I've realised is that we can also have `loaded = false` & `collection.length==0` for a second when landing on the tab. There might be a more clever way to avoid that...

melMass:
I tested it a bit and it seems to be working well!

For the creation I think it messes up with the no-route bug, to test go to https://teia-ui-pr-32.onrender.com/ROBNESS_V3

melMass (changes requested):
Looks good !

I just have the dead road issue resulting in a blank profile with "no creation":
https://teia-ui-pr-32.onrender.com/ROBNESS_V3

Should'nt we have a 404?

xat:
yes, that's a problem since forever. see for example https://www.hicetnunc.xyz/ROBNESS_V3

do we want to tackle it as part of this pr? otherwise, i would suggest that one of the designers comes up with a nice TEIA-404 image and we implement that in a separate issue.

xat:
> > i think this whole component needs heavy refactoring and would greatly benefit from a library like `useSWR`.
> 
> Yes my first commit I realised I did not understand it was a huge "if" list 😅 ..... Thanks for the fix! Another thing I've realised is that we can also have `loaded = false` & `collection.length==0` for a second when landing on the tab. There might be a more clever way to avoid that...

`useSWR` would be the more clever way, but a complete refactor of the component would be needed :-)

melMass:
> i would suggest that one of the designers comes up with a nice TEIA-404 image and we implement that in a separate issue.

Yep agreed! Let's open a dedicated issue for other devs to pickup ;)

melMass (approved):

Issue #28, opened 2022-03-20
[bug] load time on ATH sorting is high: 60s+


Baloguna16:
**Describe the bug**
While trying to sort the marketplace by ATH as per the button/link on the main page, the collection takes 60+ seconds to load without any indication that the page is still loading.

**To Reproduce**
1. Go to teia.art
2. Click on 'ATH'
3. See error/loading issue

**Expected behavior**
I would expect the load time to be faster or give some indication that the page is loading--left me thinking that I didn't click the link/that the page was broken.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Brave
 - Version 1.35.104

xat:
i think this is because of the ordering by swap.price. the price column would need an index, to make it faster. see: https://github.com/IntegroLabs/hicdex/blob/main/src/hicdex/models.py#L107

although the question is if it even makes sense to have an ATH filter in the first place? maybe we should just remove it from the ui, unless we think it brings enough value.

melMass:
> i think this is because of the ordering by swap.price. the price column would need an index, to make it faster. see: https://github.com/IntegroLabs/hicdex/blob/main/src/hicdex/models.py#L107
> 
> 
> 
> although the question is if it even makes sense to have an ATH filter in the first place? maybe we should just remove it from the ui, unless we think it brings enough value.

Yep, moon suggested removing it for what it means, I kind of agree

PR #36, opened 2022-03-23
removed ATH filter

xat:
this indirectly also solves https://github.com/teia-community/teia-ui/issues/28


melMass (approved):
Thanks! Not sure if we all agree on removing it but we can always add it back later on if really needed

PR #37, opened 2022-03-23
fixed random-filter

xat:
@melMass i know you fixed this in a better way in that refactoring-pr, but i created a separate pr for this one anyway, so that we can get the fix deployed before tackling that bigger pr.


melMass (approved):
Nice call!
For now I managed to properly rebase it 😅  I still have hope ahaha.

PR #41, opened 2022-03-23
fix: ⚡️ use github for the logos

melMass:


src/components/header/index.js
```diff
@@ -72,8 +72,8 @@ export const Header = () => {
               {/* HIC LOGO */}
               {true && (
                 <img
-                  src={`https://teia.art/logos/${context.theme}/${context.logo}`}
-                  alt="teia"
+                  src={`https://raw.githubusercontent.com/teia-community/teia-logos/main/dist/logos/${context.theme}/${context.logo}`}
```
xat: would it make sense to move this url into an env variable as well? or just store the `https://raw.githubusercontent.com/teia-community/teia-logos/main/dist/` part as env?

melMass: Makes sense! so using on env by removing 'logo.json' right?
Doing it ;)

melMass: Done in a39d78e2e8372b2c9a7aae1a2e7de852e10e0521

xat:
getting this error (although it seems to work anyway):

Bildschirmfoto 2022-03-23 um 20 20 59

melMass:
> getting this error (although it seems to work anyway):

I have the issue as well let me check locally!

melMass:
@xat Not very elegant but I think it's fixed. I set the initial value of theme to `unset`.
The initializer check if != "light" so I think it's okish...

xat (approved):
looking good!

PR #45, opened 2022-03-25
feat: ⚡️ Ukraine relief feed

### LINKED ISSUE #43: [feat] gallery/page for Ukraine fundraiser effort
**Is your feature request related to a problem? Please describe.**

Having a gallery/page for Ukraine fundraiser works would assist discovery of these collaborative works and help raise funds for the Ukraine no-war effort

**Describe the solution you'd like**

Could we please have either:

option (1): a dynamic page/view of the Ukraine collaboration art works, linked via menu/appropriate method - _preferable option as this will show current list of works dynamically_

option (2): static gallery using the galleries.json method - for this method the list of works would need to be updated/refreshed periodically so is more work operationally, to keep the list of works current. but could be an option if dynamic method is not available

**Describe alternatives you've considered**
see (2) above. other options are external galleries - NFTBiker has a page also

**Additional context**
background discussion on discord thread: https://discord.com/channels/908672304236625970/951351576839221248/954707960221827072

for option (1)
@melMass  mentioned (1) could be an option via using a graphQL “live” query

for (1), nftBiker has this query: https://discord.com/channels/908672304236625970/908710947412140082/950465754292158524
I'm not sure which fields are needed for UI to display the art, so it may need to be modified. I could try help with that if needed (did try a simpler query for just obkjt id to test (2) and that worked ok)

thanks

melMass:
* feat:⚡️ first implementation of the ukraine feed
Screenshot 2022-03-25 at 06 26 28

Closes #43

### DIFF
--- a/src/data/hicdex.js
+++ b/src/data/hicdex.js
@@ -251,7 +251,7 @@ export async function getObjktsByShare(addresses, min_shares) {
   )
 
   if (errors) {
-    console.error(errors)
+    throw errors
   }
   let objkts = []
   const contracts_info = data.hic_et_nunc_splitcontract
@@ -268,11 +268,10 @@ export async function getObjktsByShare(addresses, min_shares) {
         }
         return undefined
       }),
-      (token) => token !== undefined
+      (tokens) => tokens !== undefined
     )
 
     objkts = objkts.concat(...tokens)
   }
-  objkts = _.orderBy(objkts, ['timestamp'], ['desc'])
-  return objkts
+  return _.orderBy(objkts, ['timestamp'], ['desc'])
 }
--- a/src/data/hicdex.js
+++ b/src/data/hicdex.js
@@ -1,3 +1,5 @@
+const _ = require('lodash')
+
 export const getDipdupState = `query {
   hic_et_nunc_dipdup_state {
     hash
@@ -135,6 +137,39 @@ hic_et_nunc_shareholder(where: {holder_id: {_eq: $address}, holder_type: {_eq: "
 }
 }`
 
+export const getCollabTokensForAddressesByShare = `query GetCollabTokens($addresses: [String!]) {
+  hic_et_nunc_splitcontract(where: {shareholder: {holder_id: {_in: $addresses}}, contract: {tokens: {supply: {_gt: "0"}, _and: {is_signed: {_eq: true}}}}}) {
+    contract {
+      address
+      name
+      shares {
+        total_shares
+        shareholder {
+          holder_id
+          shares
+        }
+      }
+      tokens(order_by: {timestamp: desc}) {
+        id
+        artifact_uri
+        display_uri
+        thumbnail_uri
+        timestamp
+        mime
+        title
+        description
+        supply
+        royalties
+        creator {
+          name
+          address
+        }
+      }
+    }
+  }
+}
+`
+
 export const getManagedCollabs = `query GetManagedCollabs($address: String!) {
 hic_et_nunc_splitcontract(where: {administrator: {_eq: $address}}) {
   id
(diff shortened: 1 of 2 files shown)


melMass:
@xat It's not super clean because I could not tailor the request only in graphQL, to get only the tokens array based on the factors I used (supply >1, verified, Shares > 50% for the relief contract). The js part might surely be better but I think it's a good starting point

melMass:
@xat @AliaK77 

So what do we do? I just rebased it and added the word `Ukraine`, not sure as it shifts the tags bar in two rows (like it was before we removed ATH a few issues ago so I'm not sure)

src/data/hicdex.js
```diff
@@ -207,3 +242,37 @@ export async function fetchGraphQL(operationsDoc, operationName, variables) {
...
+
+export async function getObjktsByShare(addresses, min_shares) {
+  const { errors, data } = await fetchGraphQL(
+    getCollabTokensForAddressesByShare,
+    'GetCollabTokens',
+    { addresses }
+  )
+
+  if (errors) {
+    console.error(errors)
```
xat: maybe throw or return here, because the code that follows expects a `data` object and will fail if it's undefined.

src/data/hicdex.js
```diff
@@ -207,3 +242,37 @@ export async function fetchGraphQL(operationsDoc, operationName, variables) {
...
+      contract.shares[0].shareholder.map((shareholder) => {
+        if (
+          addresses.includes(shareholder.holder_id) &&
+          (shareholder.shares / total) * 100 >= min_shares
+        ) {
+          return contract.tokens
+        }
+        return undefined
+      }),
+      (token) => token !== undefined
```
xat: isn't this `tokens`? (plural?)

src/data/hicdex.js
```diff
@@ -207,3 +242,37 @@ export async function fetchGraphQL(operationsDoc, operationName, variables) {
...
+    )
+
+    objkts = objkts.concat(...tokens)
+  }
+  objkts = _.orderBy(objkts, ['timestamp'], ['desc'])
```
xat: not that it matters, but you could directly return here.

src/data/hicdex.js
melMass: You mean the last one? It's to filter undefined token (singular) in the tokens array

src/data/hicdex.js
melMass: Yes tbh I copied the other methods but I usually try/catch those. We could also use the graph-ql package, it's very nice, I'm actually working on a little library wrapper that use most of the request we make here, I will propose it once I'm done :)

src/data/hicdex.js
xat: yes, the last one. hmm, in the map function above you are returning `contract.tokens` (plural).

xat: at least to me, it looks like the `const tokens` from line 261 contains an array of arrays.

melMass: True there might be better, but I'm deconstructing it line 274

melMass: Or are you just referring to renaming the variable token to token_list for instance line 271? 🤣

xat: yes, it's technically correct. just the variable name `token` could be changed to `tokens`

melMass: Ahahaha! Will do those edits

src/data/hicdex.js
melMass: Hmmm how are we handling them then? I never seem to be able to hit an error (providing a wrong address for instance).
I think this is because how it's handled in fetchGraphQL
I think it's out of scope for this issue

xat (approved):
nicely done!

PR #47, opened 2022-03-25
fix: 🔥 add back the interactive symbol

### LINKED ISSUE #40: [feat] add back the </> icon for generative tokens
Multiple people asked for a clear indicator for generative tokens in the feed/thumbnails
there was this </> indicating a OBJKT is a generative one, maybe just bring that back?
currently generative tokens are indistinguishable from images in the feed

melMass:
image

Closes #40


melMass:
cc @SableRaf 🤣
https://teia-ui-pr-47.onrender.com/sableraph

src/components/media-types/html/index.js
```diff
@@ -133,7 +133,7 @@ export const HTMLComponent = (props) => {
           <div className={styles.button}>
             <Button>
               <VisuallyHidden>View</VisuallyHidden>
-              {/*               <svg
```
xat, line 136: @melMass do you know why this was commented out?

melMass: I'm going to answer on discord 😅

melMass:
After our discussion, @xat, I'm going to work on an icon cache on top of #25 😬

It's mostly there, I have to test a few different scenarios as the cache only happens in `./src/pages/display` but other page/components are using `renderMediaType` (collab for one) it needs to be added to each

xat:
looks good!

while loading all objks have the interactive icon (also the ones that are not interactive). but i think this is something we can address later.

xat (approved):
nice 👍

melMass:
> looks good!
> 
> while loading all objks have the interactive icon (also the ones that are not interactive). but i think this is something we can address later.

I'm not sure why, I think it's the way tokens load. They shift place so the icon loads before the IPFS content and it seems like it's adding it to the wrong tokens but I don't think it is as it's mimetype based

PR #52, opened 2022-03-31
feat: 🎨 proper font

melMass:
* Added various `FIXME: April` to all the edits.


melMass:
We should use CSS variables more, this makes no sense I had to edit 3 files and 5 instances

melMass:
@xat if you are around it would be nice to merge this soon! I finally added a CDN font for comic sans as some device don't have it 😅

PR #54, opened 2022-04-01
fix: 💫 revert comic sans

melMass:


melMass:
I already reverted the logos in https://github.com/teia-community/teia-logos

PR #64, opened 2022-04-08
fix: 🎨 update templates

### LINKED ISSUE #62: [feat] Better templates
The templates are a bit outdated and some people still rely on them.
It would be good to update their dependencies and make sure they work as expected.
There are bugs in the A-Frame one apparently (reported on Versum's Discord)

- [x] Update dependencies.
- [x] Fully test them
- [ ] Check hicetnunc2000 PRs as I think there was some proposed templates there

melMass:
* dependencies
* modifications when needed
* use p5.min instead of p5 (we could remove from history too later (4mb) 😬)

Closes #62


melMass:
I also made this template to maybe include in the readme? I could do that a bit later

PR #79, opened 2022-04-25
fix: ⚡️ merge n lists (both allow and restrict)

melMass:
Removes HEN list fetching

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,3 +1,5 @@
+import _ from 'lodash'
+
 let LANGUAGE = {}
 export const setLanguage = (data) => (LANGUAGE = data)
 export const getLanguage = () => LANGUAGE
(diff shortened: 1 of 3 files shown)


xat:
currently, both, the restricted-list and the permitted-list, are coming from the teia repo. does it make sense to keep these lists separated? because otherwise, this could be one single list with only the restricted accounts. the permitted accounts could just be removed from the restricted-list (since we are in control of that list).

LeonNicholls:
The separate permitted list does give us the option of overriding other lists that we might combine with the teia list for all the restrictions.

melMass:
> currently, both, the restricted-list and the permitted-list, are coming from the teia repo. does it make sense to keep these lists separated? because otherwise, this could be one single list with only the restricted accounts. the permitted accounts could just be removed from the restricted-list (since we are in control of that list).

Imo it's more modular, if we want to optin others list for instance.

xat:
makes sense 👍

xat (approved):
👍

Issue #82, opened 2022-04-28
[bug] collab mints are on the HEN marketplace once swapped, not the Teia marketplace, and don't show up on objkt


aebrer:
hello! Small bug to report regarding Teia swaps. 

Swaps on collab pieces are not showing up on objkt
example: https://objkt.com/asset/hicetnunc/717500
on Teia: https://teia.art/objkt/717500

according to MarchingSquare on objkt discord, this is because the piece is actually listed on the HEN marketplace:
![image](https://user-images.githubusercontent.com/16313135/165810672-8235320a-fa5d-423a-9386-f0949b432c34.png)

aebrer:
resolved with latest merges!

PR #83, opened 2022-04-30
Added Teia marketplace support for collab contracts

ztepler:
This commit contains changes required to migrate collab contracts from V2 marketplace to Teia marketplace. There is no back support for V2 swaps for collab contract, if this logic is required - it might be done by adding one more special swap case.

As far as I already started to work with `xat/fix/refactorings`, I decided not to switch back to main. Almost all collab swap logic was working good in this branch, but there was two exceptions: I wasn't able to pick collab in UI and looks like reswap is not working in this branch.

It would be nice to test if collab contract creation is still works: I refactored packParticipantsData function to be used with lambdas and I was not able to test it in this branch.

Also I decided to refactor createSwapCalls function a bit, because it got too sophisticated with all this proxy logic. I don't know if you accept this, but I feel it is better to split all this logic, but maybe it got a little messy. Finally, as far as reswap wasn't working in this branch, so I did not tested if it would work with all the changes I made.

### DIFF
--- a/src/components/collab/constants.js
+++ b/src/components/collab/constants.js
@@ -61,3 +61,16 @@ export const mockData = [
 export const createProxySchema = `
 (map address (pair (bool %isCore) (nat %share))))
 `
+
+export const teiaSwapSchema = `
+(pair (address %marketplaceAddress)
+  (pair %params (address %fa2)
+    (pair (nat %objkt_id)
+      (pair (nat %objkt_amount)
+        (pair (mutez %xtz_per_objkt)
+          (pair (nat %royalties) (address %creator)))))))
+`
+
+export const teiaCancelSwapSchema = `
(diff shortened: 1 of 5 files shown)


xat:
@ztepler with no back support for v2 swaps, do you mean that it is not possible to cancel v2 swaps? i suppose a lot of users want to cancel their v2 swaps in order to then swap them on teia, otherwise, the swap doesn't show up on objkt.com.

ztepler:
@xat hmm, yes, this is very good point! I will add extra case to make possible cancel V2 swaps too

aebrer:
just commenting to link this to issue #82, which this will resolve when merged!

thanks! ❤️

xat:
very nice :-)

i'll merge this now and create a new pr on top of the `xat/feature/dipdup-5` branch. then we can ask users to do some testing, before it gets merged into main.

PR #86, opened 2022-05-05
fix: ⚡️ new objkt infinite scroll

melMass:
this will be wiped once #80 is merged.
Temp fix for the feed issue.

- Properly reset the state of the feed component.
- Added offest to "new OBJKTs" to allow "loadMore"
- Switch to api.teia.art


xat (approved):
i tested it and it works as expected 👍

PR #98, opened 2022-06-03
fix: 🎨 make build/dev work OTB under powershell

melMass:


NoRulesJustFeels:
Where does craco get installed? Should it be a dev dependency in package.json?

LeonNicholls (approved):
Works fine on my Windows PC.

PR #100, opened 2022-06-03
fix: ⚡️ Jimmy two times

melMass:
Fixes  the duplicated sentence in the Swap warning message, IMO we should get rid of it entirely from the swap tab, but until this is decided here is the fixed version


merchantcoppola:
what would be a good place to have the donation address?
should have a good visibility
maybe the footer?

PR #101, opened 2022-06-06
fix: ⚡ Switch indexer (temporary)

melMass:


melMass:
As the api.teia.art is not working since the update Zir0h accepted that we temporarly switch to it

LeonNicholls (approved):

PR #102, opened 2022-06-07
feat: ⚡️ add burn transactions to history

### LINKED ISSUE #69: [feat] show burn action in history? (+bug)
**Describe the bug**
Seems that in the new UI, history doesn't show burns.
This then appears to cause swap numbers to vary, and the initial mint seems to be incorrect.

**To Reproduce**  EXAMPLE
1. Go to '...'  [715249](https://teia.art/objkt/715249) 
2. Click on history

4. See issue-   it says minted 10 / swapped 20

**Expected behaviour**
I'm assuming that 20 were minted, 20 were swapped and then 10 were burnt.
expect to see either that 10 were burned (and initial mint was 20) , or minted 10 / swapped 10 (if none were burned)

This is the only example I have evidence for, but have noticed inconsistencies in History listings before, that seem to imply burning may have happened.

Zir0h:
fixes #69
This probably still needs some work. I'm not a dev.


Zir0h:
Does anyone have a "nice" burn icon? Right now I just put X as a placeholder
2022-06-07-080059_1444x217_scrot
.

melMass:
Thanks!!
I'll have a look today

melMass:
On it!
I poked the hicathon team in case they have a proper icon. It's already working well, the only issue I see is name resolution is done using tzkt's API (thus their verification system) and not SUBJKT's

melMass:
I'm sorry @Zir0h I did not realise it would overwrite the branch in your fork 😅 .

image

Zir0h:
Thank you for the help! Looking good!

With the name resolution, you mean the sender/receiver alias?
So, I need to add a graphql query to try to translate the tz addresses to the pretty names via the holder table?

Maybe something like:
```query getHolderAlias {
  holder(where: {address: {_in: ["tz1VdRFNxALoiYPbD7LWdacZQSVcc2YudM6D"]}}) {
    address
    name
  }
}
```

And in the 'where' clause all the burners addresses.

melMass:
> I need to add a graphql query to try to translate the tz addresses to the pretty names via the holder table?

Exactly! after gathering all the 'burn transfers' to make only one call. There is already a utility function for that QL query, something like fetch_profile but don't worry too much this will all be refactored
I can look at it this afternoon!

melMass:
As I already messed up by recommiting all your commits as mine 🙈, I will reopen this PR so we can have a preview staging url

PR #107, opened 2022-06-09
hide-banned-objktcom-swaps-and-add-restricted-label

### LINKED ISSUE #93: [feat] add a "restricted" tag to restricted addresses on the "listings" tab
swaps from restricted accounts are not executable through teias UI
but it is currently unclear to the user why a swap is displayed without a price
i suggest adding a "restricted " tag to either the swap (instead of the contract info that is currently displayed like H=N, Teia, Objktcom)

or - imo even better - mark an restructed account as restricted right next to the address/SUBJKT to warn people 

or - if we want to stay minimalist - make the address name darker than the nor restricted ones

save

Theohorsmeier:
This PR addresses https://github.com/teia-community/teia-ui/issues/93 first of all, but also fixes a bug with restricted Objkt listings showing.

From this:
image

To this:
image

### DIFF
--- a/src/components/listings/index.js
+++ b/src/components/listings/index.js
@@ -4,7 +4,11 @@ import { Button, Primary, Purchase } from '../button'
 import { walletPreview } from '../../utils/string'
 import styles from './styles.module.scss'
 import { HicetnuncContext } from '../../context/HicetnuncContext'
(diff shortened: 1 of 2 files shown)


melMass (approved):
Nice it looks good!

PR #108, opened 2022-06-09
feat: ⚡️ add burn transactions to history (Zir0h)

### LINKED ISSUE #69: [feat] show burn action in history? (+bug)

melMass:
fixes #69 

resuming #102


src/data/hicdex.js
```diff
@@ -472,16 +474,66 @@ export async function fetchObjkts(ids) {
   return data.token
 }
 
+async function fetchSubjktNames(ad) {
```
xat: not that important, but why not just call the argument `addresses`?

src/data/hicdex.js
```diff
@@ -472,16 +474,66 @@ export async function fetchObjkts(ids) {
...
+  const { errors, data } = await fetchGraphQL(
+    `query SubjktNames ($addresses: [String!]) {
+    holder(where: {address: {_in: $addresses}}) {
+      address
+      name
+    }
+  }
+  `,
+    'SubjktNames',
+    { addresses: [...ad] }
```
xat: why is a new array created here?

src/data/hicdex.js
```diff
@@ -472,16 +474,66 @@ export async function fetchObjkts(ids) {
...
+        amount: item.parameter.value[0].txs[0].amount,
+      })
+    })
+  } catch (error) {
+    console.error('Problem with call to tzkt')
+    console.error(error)
+  }
+  const resolvedAddresses = await fetchSubjktNames(addressesToResolve)
+
+  _.each(result.transfers, (transfer) => {
```
xat: using `result.transfers.forEach` or `result.transfers.map` would be more consistent with `burn_operations.forEach`.

src/data/hicdex.js
```diff
@@ -472,16 +474,66 @@ export async function fetchObjkts(ids) {
...
+      if (addressesToResolve.indexOf(item.sender.address) === -1) {
+        addressesToResolve.push(item.sender.address)
+      }
+
+      result.transfers.push({
+        timestamp: item.timestamp,
+        ophash: item.hash,
+        sender: item.sender.address,
+        opid: item.id,
+        amount: item.parameter.value[0].txs[0].amount,
```
xat: the txs array could have multiple entries. the burn transfer is not necessarily the first element in the array.

src/pages/objkt-display/tabs/history.js
```diff
@@ -4,17 +4,35 @@ import { Primary } from '@components/button'
...
 
 import { IconCache } from '@utils/with-icon'
-import TradeIcon from '@icons/trade'
-import MintedIcon from '@icons/minted'
-import SwapIcon from '@icons/swap'
+import { TradeIcon, MintedIcon, SwapIcon, BurnIcon } from '@icons'
+// import MintedIcon from '@icons/minted'
+// import SwapIcon from '@icons/swap'
+
+class OperationType {
```
xat: i think it would be simpler and easier to read to just use regular constants here:

```
const OPERATION_SWAP = 'SWAP';
const OPERATION_TRANSFER = 'TRANSFER';
const OPERATION_TRADE = 'TRADE';
```

xat (commented):
added a few small comments.

src/data/hicdex.js
melMass: Indeed, I was having issues and though JS was turning arrays into string when there was only 1 value, it's not the case

src/data/hicdex.js
melMass: I'm pretty sure it is given the endpoint call

src/pages/objkt-display/tabs/history.js
melMass: I was trying to do enums and discovered there is none in JS.
Can't we just use strings directly then?

src/data/hicdex.js
melMass: Indeed, btw is there a reason to use lodash's one? 
I'm using lodash everywhere personally, since you show me composition

xat: the lodash one also accepts objects as input and probably doesn't complain if the input is undefined. but in general, it's nicer to just use the javascript built-in functions (at least in my opinion).

src/data/hicdex.js
xat: i also think so, but you can perform multiple transfers in one single call.

src/data/hicdex.js
melMass: Vanilla if I don't make use of composition, makes sense

src/pages/objkt-display/tabs/history.js
xat: i think that also would be fine in this case. constants guard a bit against potential typos using strings (but indeed not that important in this case)

src/data/hicdex.js
melMass: @xat So I used lodash 😅 tell me if it's what you had in mind

src/data/hicdex.js
```diff
@@ -472,16 +474,66 @@ export async function fetchObjkts(ids) {
...
+      if (addressesToResolve.indexOf(item.sender.address) === -1) {
+        addressesToResolve.push(item.sender.address)
+      }
+
+      result.transfers.push({
+        timestamp: item.timestamp,
+        ophash: item.hash,
+        sender: item.sender.address,
+        opid: item.id,
+        amount: _(item.parameter.value[0].txs).sumBy('amount'),
```
xat: i think you need to filter out the transfers where the `_to` property is not set to the burn address.

here is a random example of how the transfer entrypoint can be called:

```
parameter: {
    entrypoint: 'transfer',
    value: [
        {
            txs: [
                {
                    to_: 'tz1ejpAPJZAe2boiKUGwb6cq5dckTZ8hXn2A',
                    amount: '1',
                    token_id: '58486',
                },
                {
                    to_: 'tz1UBZUkXpKGhYsP5KtzDNqLLchwF4uHrGjw',
                    amount: '10',
                    token_id: '58487',
                },
            ],
            from_: 'KT1Hkg5qeNhfwpKW4fXvq7HGZB9z2EnmCCA9',
        },
    ],
},
```

xat (approved):
perfect 👍

PR #109, opened 2022-06-10
quick fix: adjusting transfer error text

### LINKED ISSUE #42: [support] What's the use of the "Transfer" tab on OBJKTs in my wallet?
I've come across it today and though how convenient. It doesn't appear to be for FA2 transfers?

Thanks!

merchantcoppola:
since i dont know how to hide this tab when not in a collab i adjusted the warning text to explain that the transfer tab is only for sending objkts out of a collab contract. to avoid confusion


melMass:
Thanks, is there an issue for that? It should be fairly easy to properly implement

melMass (approved):

PR #104, opened 2022-06-08
Minting metadata improvements 

LeonNicholls:
Based on https://docs.google.com/document/d/1R3rxuJfpCBoh52wDsYpEmY_iuH8sd6l17oh4nk7jlRE/edit?usp=sharing

New fields still need to be added to the GUI.

PR #113, opened 2022-06-16
feat: ⚡️ minting metadata improvements (Metadatas + UI)

### LINKED ISSUE #115: [bug] clear fields should work even when there is no local storage
Reported on the discord.
On the mint page, If one does not have any fields in local storage it's not clearing the UI

melMass:
UI part of #104 (mainly separated to get our staging link)


src/data/ipfs.js
```diff
@@ -14,106 +16,77 @@ const infuraUrl = 'https://ipfs.infura.io:5001'
...
+    }
+  }
 
   return await uploadMetadataFile({
     name,
     description,
     tags,
     cid,
     address,
-    mimeType,
```
melMass, line 114: Is there a reason to remove mimetypes?

melMass: Ah yes you resolve from extensions later on, I'm just refactoring this a bit

melMass:
I just went on integrating the UI and components.
Maybe we should use this PR to clean the mint file a bit

src/constants.js
```diff
@@ -166,3 +166,278 @@ export const MAIN_MARKETPLACE_CONTRACT = MARKETPLACE_CONTRACT_TEIA // the one th
...
+  'cc-by-4.0': 'CC-BY-4.0 (Attribution)',
+  'cc-by-sa-4.0': 'CC BY-SA 4.0 (Attribution ShareAlike)',
+  'cc-by-nd-4.0': 'CC BY-ND 4.0 (Attribution-NoDerivs)',
+  'cc-by-nc-4.0': 'CC BY-NC 4.0 (Attribution-NonCommercial)',
+  'cc-by-nc-sa-4.0': 'CC BY-NC-SA 4.0 (Attribution-NonCommercial-ShareAlike)',
+  'cc-by-nc-nd-4.0': 'CC BY-NC-ND 4.0 (Attribution-NonCommercial-NoDerivs)',
+  custom: 'Custom (Specify)',
+}
+
+export const LANGUAGES = {
```
melMass, line 214: @NoRulesJustFeels Do you have a better source 😅 . 
I'm not sure Country ISO code is what was in the metadata proposal, but I improvised just to fill it for now

LeonNicholls: This is the standard for the language based on the TZIP-21 spec: https://datatracker.ietf.org/doc/html/rfc1766
But it seems there have been other standards that have replaced this. Based on a search, I found this list: https://www.loc.gov/standards/iso639-2/php/code_list.php

melMass:
Ok I'll leave it there for today, remaining:

- [x] Sort languages by language name rather than 'iso' code
- [x] Validate that rightsURI is a proper URL (must start with "https://" or "ipfs://"
- [x] Styling: Mixing the mint field animations with the new components or find a workaround that looks good (@SlinkousArt if you want to look into it too)
- [x] Clear fields does not work on "selects"
- [x] Product testers round

melMass:
~~I should most probably revert b23cd16ad679bdad05fe7090baca10c3f6146ab5, just commenting to not forget.~~

Still needs testing but 26b9570bc81593e8c8cb4b8133afd22158c2de8e solves this

xat:
the ui works nice 👍 

was wondering if we should maybe make the title mandatory? or are there cases where tokens are minted without title?

src/components/input/index.js
```diff
@@ -14,38 +14,70 @@ export const Input = ({
...
       <input
         type={type}
         placeholder={placeholder}
-        name={name}
+        name={name || label}
         min={min}
         max={max}
         maxLength={maxlength}
-        defaultValue={value}
+        defaultValue={defaultValue === null ? '' : defaultValue}
```
xat, line 33: out of interest, why is defaultValue needed?

src/components/input/index.js
```diff
@@ -14,38 +14,70 @@ export const Input = ({
...
+        value={value === null ? '' : value === undefined ? '' : value}
```
xat: this might be easier to read:

```
value={(value === null || value === undefined) ? '' : value}

File: src/components/input/index.js (hunk 3 above)
Comment (melMass) [reply src/components/input/index.js:33]: This is one of the things I had the most issue with.
Basically it's the controlled vs uncontrolled input thing I linked in b23cd16ad679bdad05fe7090baca10c3f6146ab5 since now, the Input component was only exposing `value` but was used as `defaultValue`... So I unshadowed (26b9570bc81593e8c8cb4b8133afd22158c2de8e) that and kept using uncontrolled/controlled inputs where needed, it would have been easier to use a form library as you suggested 😅

File: src/components/preview/index.js [hunk 5]
```diff
@@ -28,9 +41,43 @@ export const Preview = ({ title, description, mimeType, previewUri, tags }) => {
...
+        <div className={styles.title}>
+          <strong>Title:</strong>
+          {title}
+        </div>
+        <div className={styles.description}>
+          <strong>Description:</strong>
+          {description}
+        </div>
+
+        <div className={styles.description}>
```
xat: maybe rename `styles.description` to something more general, like `styles.field` (if it's not just used for the description)

src/components/preview/index.js
```diff
@@ -28,9 +41,43 @@ export const Preview = ({ title, description, mimeType, previewUri, tags }) => {
...
+            <strong>Attributes:</strong>
+            {nsfw && (
+              <span className={styles.label} title={'This artwork is NSFW'}>
+                NSFW
+              </span>
+            )}
+            {photosensitiveSeizureWarning && (
+              <span
+                className={styles.label}
+                title={'This artwork is can cause seizures'}
```
xat: the word `is` can be removed

src/components/select/index.js
```diff
@@ -0,0 +1,73 @@
...
+    backgroundColor: 'var(--background-color)',
+    border: 'none',
+    // none of react-select's styles are passed to <Control />
+    // width: 200,
+  }),
+  menu: (provided, state) => ({
+    ...provided,
+    backgroundColor: 'var(--background-color)',
+  }),
+  clearIndicator: (provided, state) => ({
```
xat: can probably be removed, because no styles get changed

melMass: Yes that's part of the leftovers I planned to remove, I'll do a pass with your edits and these later today 🚀

src/constants.js
```diff
@@ -166,3 +166,169 @@ export const MAIN_MARKETPLACE_CONTRACT = MARKETPLACE_CONTRACT_TEIA // the one th
...
+  'cc-by-4.0': 'CC-BY-4.0 (Attribution)',
+  'cc-by-sa-4.0': 'CC BY-SA 4.0 (Attribution ShareAlike)',
+  'cc-by-nd-4.0': 'CC BY-ND 4.0 (Attribution-NoDerivs)',
+  'cc-by-nc-4.0': 'CC BY-NC 4.0 (Attribution-NonCommercial)',
+  'cc-by-nc-sa-4.0': 'CC BY-NC-SA 4.0 (Attribution-NonCommercial-ShareAlike)',
+  'cc-by-nc-nd-4.0': 'CC BY-NC-ND 4.0 (Attribution-NonCommercial-NoDerivs)',
+  custom: 'Custom (Specify below)',
+}
+
+export const LICENSE_TYPES_OPTIONS = Object.keys(LICENSE_TYPES).map((k) => {
```
xat, line 207: afaik javascript does not guarantee a specific order of the object keys, but in practice, browsers order them as you would expect ('none' first).

melMass: I did not know that 🙈 ! I guess we could ensure it with and `_.orderBy` but I tested a few browsers and it seemed consistent indeed (browser odering them)

melMass:
> was wondering if we should maybe make the title mandatory? or are there cases where tokens are minted without title?

Yes @merchantcoppola just wondered the same in #121, I have no opinion, but we could!

src/constants.js
xat: yes, no need to change it, in practice it's not a problem. a Map guarantees the order btw, but that is clunky to use :)

xat (approved):

melMass:
I'm squash merging it, but won't delete the branch just yet just in case ;)

PR #117, opened 2022-06-25
fix: ⚡️ prevent restricted swaps from being the main swap

### LINKED ISSUE #116: [bug] lowest price still collectable from banned accounts
**Summary**
As a collector, I am still allowed to collect from banned accounts using the HENv2 contract when the swapped price is the lowest price.

**Description**
Example
- https://teia.art/objkt/45586
- Banned account is https://teia.art/hen
- Banned account listing is 0.15 and is still shown on the main collect button, but is restricted on the "Listings" tab
- Any objkt listed as the lowest price by a banned account can still be collected by users.

**Screenshots**
Screen Shot 2022-06-24 at 8 31 44 AM

**To Reproduce**
1. Go to https://teia.art/objkt/45586
2. Click on "Collect for 0.15 tez"
3. User is still allowed to collect from the "hen" banned account in "Listings"

**Expected behavior**
Listings for banned accounts should not be shown when they are the lowest swap. Using the example from above, the 1 tez listing should be shown instead of the 0.15 tez listing.

**Environment**
 - OS: Mac OSX
 - Browser Chrome
 - Version 103.x

Theohorsmeier:
Fixes https://github.com/teia-community/teia-ui/issues/116

image

### DIFF
--- a/src/components/item-info/index.js
+++ b/src/components/item-info/index.js
@@ -33,7 +33,18 @@ export const ItemInfo = ({
   if (isDetailView) {
     // TODO: subtract burned pieces from total
     const total = supply
-    const editionsForSale = sum(listings.map(({ amount_left }) => amount_left))
+    const editionsForSale = sum(
+      listings
+        .filter(
+          (listing) =>
+            !ban.includes(
+              listing.seller_address
+                ? listing.seller_address
+                : listing.creator.address
+            )
+        )
+        .map(({ amount_left }) => amount_left)
+    )
     const ed = editionsForSale || 'X'
     let purchaseButton = null
 
--- a/src/components/item-info/index.js
+++ b/src/components/item-info/index.js
@@ -23,6 +23,7 @@ export const ItemInfo = ({
   supply,
   isDetailView,
   restricted,
+  ban,
 }) => {
   const { syncTaquito, collect, fulfillObjktcomAsk, curate, acc } =
     useContext(HicetnuncContext)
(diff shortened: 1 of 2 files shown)


Theohorsmeier:
Pushed to this a second time because restricted objkts did not have the ban variable, took it out of the if/else scope

melMass:
@Theohorsmeier Hey Theo, sorry I couldn't look at it before now, is there still the issue when the only swap is from a banned account?

Theohorsmeier:
Hey @melMass , I think in that case the only problem is that the number of available editions is wrong:

image

https://teia.art/objkt/187701

melMass:
> 187701

I think it's not too heavy to filter there too. I just pushed an edit

---
**edit**: As there is no staging link:
image

src/components/item-info/index.js
```diff
@@ -33,7 +33,18 @@ export const ItemInfo = ({
...
     // TODO: subtract burned pieces from total
     const total = supply
-    const editionsForSale = sum(listings.map(({ amount_left }) => amount_left))
+    const editionsForSale = sum(
+      listings
+        .filter(
+          (listing) =>
+            !ban.includes(
+              listing.seller_address
+                ? listing.seller_address
```
melMass, line 42: objkt.com swaps

src/components/item-info/index.js
```diff
@@ -33,7 +33,18 @@ export const ItemInfo = ({
...
+                : listing.creator.address
```
melMass, line 43: teia swaps

melMass (approved):

PR #120, opened 2022-06-28
Some padding/margins

SlinkousArt:
Padding above labels is larger, padding below is consistent.


melMass:
I think you forgot to push your last edits no?
I'm closing it in favor or #113, check my last edits, maybe it can be done better, I honestly am pretty bad with CSS and from what I understand there is no "standard" used accross the site

PR #125, opened 2022-07-19
fix: ⚡️ husky on windows

melMass:
The arguments where not parsed properly on windows.
Using the package.json config instead works!


melMass:
Hmm no it actually is not running anymore with this config, but there is a windows note on the doc, I'll check it out

melMass:
I'm merging it as I'm making branches from main on windows and it's a pain

PR #126, opened 2022-07-19
feat: 💄 modernize the github templates

melMass:
Live preview here:
- **bug report:**  https://github.com/teia-community/teia-ui/blob/mel/fix/github_files/.github/ISSUE_TEMPLATE/bug_report.yml


melMass:
Merging for now, it basically adds a better UI for user reports, and set a default title etc...
This should make issues a bit cleaner OOB

PR #128, opened 2022-07-19
Fix mint transaction handling and UI feedback

LeonNicholls:


melMass:
Nice, I'll test it more locally today

melMass (approved):
Very nice!

PR #129, opened 2022-07-20
Fix metadata boolean

LeonNicholls:


melMass:
Approving it, just pinging @Zir0h in case it matters for the indexer 😬

melMass (approved):

Issue #110, opened 2022-06-10
[bug] Infinite scroll on "new OBJKTs" feed is broken


merchantcoppola:
basically what it says in the headline: 
The infinite scroll of the "new OBJKTs" feed doesnt work, stopy after 8 or so objkts

melMass:
Not really after 8 but it's indeed stopping after ~30-40.
I still haven't looked into it but it should be fairly obvious to spot with a debugger

Zir0h:
I think it stops when you encounter a large amount of restricted objects, but I haven't been able to look further.

melMass:
> I think it stops when you encounter a large amount of restricted objects, but I haven't been able too look further.

Interesting thanks! I will look at it tonight with that in mind

xat:
i can also have a look at this as part of the refactoring branch. there i changed the feed logic quite a bit.

melMass:
@xat Did you nailed the issue in the refacto? Should I maybe look into fixing it sooner than the PR merge?

xat:
yeah, it's fixed in the refactoring branch, but that might indeed still take a bit of time until it's ready to be merged.

melMass:
@xat @jagracar ... I just have a doubt, I can't see the error anymore, can you confirm? was it just not deployed when this issue was filed?

jagracar:
It's something that doesn't happen always. It might be connected with some special tokens. That's what makes this issue tricky. It happened again one week ago and since then it didn't happen again, but i would be surprised if it's solved.

jagracar:
I will try to let you know when I see it.

jagracar:
Happening now @melMass 
The last visible OBJKT is 763828

melMass:
> It might be connected with some special tokens.

That's my guess too!
I'll try to pinpoint it!

Zir0h:
It happens on recent sales too, stuck at 680282.

Probably because an objkt was very popular recently:
![image](https://user-images.githubusercontent.com/546070/182033609-141bedc7-2a06-4b0b-94fe-7e435dfedbe9.png)
748249 in this case

//edit: this looks indeed like it's fixed in #80 so nevermind! :sweat_smile: 

#80 seems stuck on 762761 for new objkt feed

melMass:
Interesting, it's still stuck on 762761 an hour later

jagracar:
I think it remains always stuck. At some moment is so old that you don't notice anymore, because you need to scroll down a lot

[...]

PR #140, opened 2022-07-28
feat: ✨ Use new moderation lists

LeonNicholls:
Support new moderation lists:
- nsfw.json: track nsfw OBJKTs that don't have the metadata set
- under_review.json: flag accounts that need to be reviewed by moderation team
- ignore_uri.json: IPFS URIs to ignore for potential terms and conditions violations


src/components/feed-item/index.js
```diff
@@ -20,6 +22,7 @@ export const FeedItem = (props) => {
             creator: props.creator_id,
             objkt: String(props.id),
             displayView: true,
+            blur: nsfwList.includes(props.id.toString())
```
melMass: Not tested yet but just looking at the edits can't the blur classname be passed one level up, on the container?

src/pages/mint/index.js
```diff
@@ -358,6 +358,14 @@ export const Mint = () => {
 
     const uri0 = `ipfs://${hashv0}`
     const uri1 = `ipfs://${hashv1}`
+
+    // Ignore IPFS URI's that are in the ignore list; they can be minted multiple times
+    const ignoreUriList = getIgnoreUriList()
+    if (ignoreUriList.includes(uri0) || ignoreUriList.includes(uri1)) {
+      console.log('ignore uri')
```
melMass: remaining console logs :)

Btw how is this list generated?

src/pages/objkt-display/index.js
```diff
@@ -156,14 +163,13 @@ export const ObjktDisplay = () => {
...
-                  <div
-                    style={{
-                      color: 'white',
-                      background: 'black',
-                      textAlign: 'center',
-                    }}
-                  >
-                    Restricted OBJKT
+                  <div className={styles.restricted}>
+                    Restricted OBJKT. Contact the Teia moderators on <a href="https://discord.gg/TKeybhYhNe" target="_blank" rel="noreferrer">Discord</a> to resolve the status. See the <a href="https://github.com/teia-community/teia-docs/wiki/Core-Values-Code-of-Conduct-Terms-and-Conditions#3-terms-and-conditions---account-restrictions" target="_blank" rel="noreferrer">Teia Terms and Conditions</a>.
```
melMass: Might be done in a separate PR, but if possible it would be better to use an "external" file for texts

src/pages/objkt-display/styles.module.scss
```diff
@@ -119,124 +96,120 @@
...
-  .history__ed,
-  .history__price,
-  .history__date {
-    display: block;
-  }
-
-  @include respond-to('desktop') {
-    display: none;
-  }
+    display: flex;
```
melMass, line 180: Did husky worked on your end? There is a lot of formatting edits

melMass:
I added a few comments besides that it seems all good, I will test it later 👍

src/pages/mint/index.js
LeonNicholls: I thought in this case having something in the console for this case might be useful.
This list is manually updated with IPFS URI's that we want to ignore.

src/pages/objkt-display/index.js
LeonNicholls: Agreed. Seems like a lot of text in the GUI at the moment.

src/pages/objkt-display/styles.module.scss
LeonNicholls: Wasn't aware of the husky step. Don't think I did anything like that before. What do I do now?

src/components/feed-item/index.js
LeonNicholls: Not exactly sure what you mean by that.

melMass: Instead of passing down `blur` 2/3 levels, just adding the classname to the container in this file

src/pages/objkt-display/styles.module.scss
melMass: It should autoinstall on npm install (the prepare script runs just after) the first time and its job is to run cmd on git events, we use it for commits, it lint/formats staged files to avoid this. I will check it maybe it actually needed those as it's only ran on staged files, but we ran it over the whole codebase once when we introduced husky so not sure.

LeonNicholls: I'm not sure if it was automatically installed. Never saw any husky errors or anything that looked like a pre-commit trigger. This is what I got with the prepare target:
npm run prepare

> teia-ui@0.1.0 prepare
> husky install

husky - Git hooks installed

src/components/feed-item/index.js
LeonNicholls: Oh. Ok. Originally, I thought that each of the different kinds of media might have a different kind of "blur" and even different levels of blur; that is why I passed it down to each component. I ended up doing the same for each kind of media, but still not sure if that is the best to go with. I could imagine that some formats might not even do a CSS blur but black out the content.

src/pages/mint/index.js
melMass: IMO there should be 0 log on a website, console.debug is meant for that if really needed.
After the refacto I wanted to convert every logs on the site (removing the useless ones)

src/pages/objkt-display/styles.module.scss
melMass: This should mean it's properly running, I'll test locally from main making the same edits to see the behaviour on my end

src/components/feed-item/index.js
melMass: This can be done with css only I think by using a class per media or something like that, I just feel that a visual feature should not be passed down that much components, maintenance wise it's not ideal

src/pages/mint/index.js
LeonNicholls: Will remove.

src/components/feed-item/index.js
LeonNicholls: Ok. I thought that "displayView' was already an example of a visual feature being passed down, so I followed that pattern. But I will change to apply the blur to the top level element since they end up being a CSS filter for all of them (at least for now).

src/pages/objkt-display/styles.module.scss
LeonNicholls: Interesting. Just did a commit and now I'm seeing this (don't think I've seen this before):
ubuntupc:/hen/teia-ui$ git commit -m "Move blur to top level"
🔍  Finding changed files since git revision f1dd7672.
🎯  Found 22 changed files.
✍️  Fixing up src/components/feed-item/index.js.
✍️  Fixing up src/components/feed-item/styles.module.scss.
✍️  Fixing up src/components/media-types/video/style.css.
✍️  Fixing up src/pages/display/index.js.
✍️  Fixing up src/pages/display/styles.module.scss.
✍️  Fixing up src/pages/mint/index.js.
✅  Everything is awesome!
[new-moderation-lists 96b80887] Move blur to top level
 23 files changed, 625 insertions(+), 590 deletions(-)
 rewrite src/components/media-types/audio/styles.module.scss (84%)
 rewrite src/components/media-types/glb/styles.module.scss (85%)
 rewrite src/components/media-types/html/styles.module.scss (82%)
 rewrite src/components/media-types/image/styles.module.scss (74%)
 delete mode 100644 src/components/media-types/md/styles.module.scss
 rewrite src/components/media-types/pdf/styles.module.scss (78%)
 rewrite src/components/media-types/vector/styles.module.scss (78%)
 rewrite src/pages/display/styles.module.scss (78%)

melMass:
I know I said it a few times already @jagracar but I think the last commit solves #110!
I'm going to merge it now, to see on the staging url

melMass (approved):
Great!

PR #142, opened 2022-07-30
fix: 🎨 use markdown for the about page

melMass:
By quickly checking I did not found any i18n library that
uses markdown but IMO the best approach would be a mix
of markdown and json.
Json for single words, short sentences and markdown for more
"complex" stuff.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,7 +46,7 @@
         "screenfull": "^5.1.0"
       },
       "devDependencies": {
-        "@craco/craco": "^7.0.0-alpha.3",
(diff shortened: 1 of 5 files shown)


melMass:
IMO this is good to go as i18n should be handled separately. We can rethink the logic then

xat:
tested, works as expected 👍 

for butiq, i'm using this markdown library btw: https://markdoc.io/
(just as info, not suggesting to switch :) )

xat (approved):
works as expected

PR #144, opened 2022-07-31
[feat] Tests

melMass:
- feat: 🐛 add playwirght dep and "boilerplate"


melMass:
@xat If you have ideas for tests that could apply to teia, shoot! I'm trying to come up with a few automatic tests (jest + playwright).
For now it's simply scrolls the feed for 60 sec but I'm not checking anything yet

xat:
@melMass personally i would just test logic-heavy functions with unit tests.

testing that includes scrolling etc. is very fragile and will need a lot of maintenance. imo, adding these kinds of tests only makes sense if we are very sure that they provide enough value.

PR #145, opened 2022-07-31
fix: ⚡️ use teia repo for html previews

melMass:
I meant to look into it before.
I've enabled pages on the repo and tested it.


melMass:
Merging as I need it for #146 and it works on it's own.

Issue #150, opened 2022-08-09
[feat] Use the nft.storage IPFS gateway


LeonNicholls:
### Describe the problem

Once we use the nft.storage IPFS service, we should consider using their IPFS gateway: https://nft.storage/docs/concepts/gateways/

### Describe the solution you'd like

Use https://nftstorage.link/ipfs/* for loading IPFS files

Zir0h:
The only issue is the 200req/minute rate limiting:
https://nft.storage/docs/concepts/gateways/#rate-limits

It's quite easy to reach, you only have to scroll through your collection for example (I've tested this gateway on .rocks)

melMass:
>  200req/minute 

Ouch, yes indeed, even if I think there is  a lot of room for optimization on our side to reduce the requests, 200/mn is prone to fails... 

I will add the option to switch to it without using it yet

LeonNicholls:
We might want to ask nft.storage if they can increase the rate limit for requests from teia.art

Zir0h:
Well, nft.storage doesn't really know where the requests are coming from I think.

We could self host it, there's a guide on their github:
https://github.com/nftstorage/nftstorage.link/blob/main/packages/edge-gateway/README.md
I think that's the only way to circumvent the ratelimiting.

melMass:
AFAIK rate limiting is done per IP on the client side, I don't know the exact challenges/costs of self hosting a gateway, but IMO focus should be put on how to fetch less data (by fixing UI code), and how to better serve tokens (by using some CDN/Caching of cover and artifacts)

LeonNicholls:
@dchoi27 is the edge gateway the only way to avoid the rate-limiting?

dchoi27:
^ hm not sure if i understand - rate limits for the edge gateway as correctly stated above is 200 req / min. I'd advise not hosting your own gateway tbh - it's a lot of work.

One thing you might consider is using the SuperHot feature of https://nftstorage.link/ (scroll down a bit on the page) which can perma-cache the content you care about on the edge server. Then we could alleviate rate limits coming from teia.art (the rate limits are there to protect the underlying infrastructure we use to get data from the IPFS network, but if we know the data is already cached, it reduces concerns of exposing the underlying infra from being DDoS'd). 

if interested, I can get your NFT.Storage account access to the SuperHot API feature for you all to play around with! cc @vasco-santos

Zir0h:
[...]

PR #159, opened 2022-08-12
[feat] NFT.storage continue...

### LINKED ISSUE #5: [feat] add error reporting when the SUBJKT is already registered
would be cool to have a clearer error mssg when trying to pic a profile name in "seetings" that has already been chosen
currently the only thing that happens is that "transaction fails"

### LINKED ISSUE #158: [bug] there's a call to infura ipfs left in the config section
### Describe the bug

It's impossible to update the Teia profile at the moment. People are unable to save and they're getting a 401 in the web dev console.

image

Possibly because of this?
src/pages/config/index.js:const infuraUrl = 'https://ipfs.infura.io:5001'
src/pages/config/index.js:const ipfs = create(infuraUrl)

### Reproduction

1. go to your profile
2. make a change
3. save

### Expected behavior

Profile is updated

### Platform and versions

```shell
- Linux
- Firefox
- 103.0.2
```

### Console output

_No response_

### Additional context

_No response_

### LINKED ISSUE #162: [bug] collab settings change doesn't prompt wallet
### Describe the bug

When trying to change the username and description for a collab contract, when I click `Save Profile` nothing happens. This happens regardless of whether or not an image is being uploaded.

### Reproduction

1. make a new collab contract (eg. https://teia.art/tz/KT1MARbF9KvMjGJWNX3j3bVdkatUjWbD4zU1)
2. "log in" to it, via the collaborate menu
3. go to settings
4. fill out new profile information
5. press "Save Profile"
6. nothing happens

### Expected behavior

When pressing `Save Profile`, a prompt to sign in your wallet should come up, and then the changes will be uploaded and saved.

### Platform and versions

```shell
- OS: Windows 11
- Browser: tried Firefox and Chrome
- tested teia.art and teia.rocks
```

### Console output

```shell
XHRPOSThttps://ipfs.infura.io:5001/api/v0/add?stream-channels=true&progress=false
[HTTP/1.1 401 Unauthorized 39ms]

	
POST
	https://ipfs.infura.io:5001/api/v0/add?stream-channels=true&progress=false
Status
401
Unauthorized
VersionHTTP/1.1
Transferred360 B (20 B size)
Referrer Policystrict-origin-when-cross-origin
Request PriorityHighest

    	
    Access-Control-Allow-Credentials
    	true
    Access-Control-Allow-Origin
    	https://teia.art
    Content-Length
    	20
    Content-Type
    	text/plain; charset=utf-8
    Date
    	Tue, 16 Aug 2022 15:17:18 GMT
    Vary
    	Origin
    Www-Authenticate
    	Basic realm="Project ID is required"
    X-Content-Type-Options
    	nosniff
    X-Robots-Tag
    	noindex
    	
    Accept
    	*/*
    Accept-Encoding
    	gzip, deflate, br
    Accept-Language
    	en-US,en;q=0.5
    Connection
    	keep-alive
    Content-Length
    	259
    Content-Type
    	multipart/form-data; boundary=---------------------------100955003518257656423327817725
    DNT
    	1
    Host
    	ipfs.infura.io:5001
    Origin
    	https://teia.art
    Referer
    	https://teia.art/
    Sec-Fetch-Dest
    	empty
    Sec-Fetch-Mode
    	cors
    Sec-Fetch-Site
    	cross-site
    User-Agent
    	Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:103.0) Gecko/20100101 Firefox/103.0
```

### Additional context

_No response_

melMass:
Fixing issues due to the IPFS switch and older unreported ones:

- [x] Cannot edit profile
   - [x] Works but there is absolutely no feedback...
- [x] MD preview / styling -> fixed displayURi issue, styling should be done is a dedicated PR
- [x] Subfolders for generative tokens (linked to https://github.com/teia-community/ipfs-upload-proxy/pull/3)
- [x] ~~Clean the folders before upload? (macOS creates .DS_Store files and `._.` prefixed files, windows has ThumbDB), might be "dangerous" to alter the archive~~


melMass:
I was about to remove every instance of IPFS as we use various gateways.
@NoRulesJustFeels while doing that I wonder if this won't make reaching gateways limits faster?

For now I just changed the SUBJKT icon (was DWEB -> IPFS.io), keeping the current gateway used by others, we can switch easily once we decide

LeonNicholls:
As part of the discussion on #150, there are various IPFS options to consider. We should probably spend some time testing those when we have access. We might even consider some kind of failover solution for the client if the rate limit is reached for a particular gateway.

melMass:
> As part of the discussion on #150, there are various IPFS options to consider. We should probably spend some time testing those when we have access. We might even consider some kind of failover solution for the client if the rate limit is reached for a particular gateway.

I was not up to date with that issue! Awesome

melMass:
## Added a few things:

- **Check the balance on the mint page**: This displays if you have less than 0.15Tz (6Tz for the screenshot)
	image
	
- **Check for existing SUBJKT**: This is an ongoing issue people had as there is no check before making the contract call so it only failed in Temple/Kukai not before. Now it's showing the owner of the name:
	image

- **Replaced alert() with the context feedback**:
	image


- All IPFS **CIDs** and **Hashes** now use dedicated util functions shared across to convert them to http gateway links. The second argument to specify the gateway is optional and by default uses a new dedicated ENV Variable (`REACT_APP_IPFS_DEFAULT_GATEWAY`). A few different gateways (DWeb, Cloudflare, Ipfs.io, Infura) where used before.

melMass:
Ah ok... The CI is not building to the stage URL since quite a few commits, related to this somehow even if we use npm everywhere
https://github.com/dilanx/craco/issues/435

jsconfig.json
```diff
@@ -1,14 +1,18 @@
 {
```
xat, line 1: out of interest: why the changes here?

src/components/collab/show/CollabDisplay.js
```diff
@@ -88,13 +89,9 @@ export const CollabDisplay = () => {
...
 
       if (metadataFile) {
-        axios
-          .get(
-            `https://cloudflare-ipfs.com/ipfs/${metadataFile.split('//')[1]}`
-          )
-          .then(({ data }) => {
-            setLogo(data.identicon)
-          })
+        axios.get(HashToURL(metadataFile)).then(({ data }) => {
```
xat: i think it would be better to use `await` here, otherwise errors won't be propagated to the `catch()` function that comes after `fetchMeta()`

src/components/media-types/html/index.js
```diff
@@ -54,24 +54,31 @@ export const HTMLComponent = (props) => {
...
-    }
-    setValidHTML(result.valid)
 
-    unpacking.current = false
+    try {
+      const filesArr = await prepareFilesFromZIP(buffer)
+      const files = {}
+      filesArr.forEach((f) => {
+        files[f.path] = f.blob
+      })
```
xat: not important (it's also not new code), but Array.reduce comes in handy in these cases:

```JavaScript
const files = filesArr.reduce((memo, f) => ({ ...memo, [f.path]: f.blob }), {});
```

src/components/media-types/index.js
```diff
@@ -93,8 +59,8 @@ export const renderMediaType = ({
 
   displayView,
 }) => {
-  let parsedArtifactUri
-  let parsedDisplayUri
+  const parsedArtifactUri = artifactUri ? HashToURL(artifactUri) : ''
```
xat, line 62: i think the `HashToURL` function needs a second argument.

src/components/media-types/video/index.js
```diff
@@ -49,37 +47,33 @@ export const VideoComponent = ({
...
+        playsInline
+        muted
+        loop
+        controls={interactive}
+        src={preview ? previewUri : artifactUri}
+        poster={displayUri}
+      />
+    </div>
+  ) : (
+    <>
```
xat: not important, but the `<></>` wrapping is not needed (i know, it was there before)

src/constants.js
```diff
@@ -90,32 +90,31 @@ export const MINT_FILESIZE = 100
...
-  SVG: 'image/svg+xml',
-  TIFF: 'image/tiff',
-  WEBP: 'image/webp',
-  MP4: 'video/mp4',
-  OGV: 'video/ogg',
-  QUICKTIME: 'video/quicktime',
-  WEBM: 'video/webm',
   GLB: 'model/gltf-binary',
   GLTF: 'model/gltf+json',
+  IPFS_DIRECTORY: 'application/x-directory',
```
xat: maybe just `DIRECTORY`, to make it consitent with the other mimetype constants.

src/pages/config/index.js
```diff
@@ -113,29 +128,112 @@ export class Config extends Component {
...
       const [file] = this.state.selectedFile
 
+      this.context.setFeedback({
+        message: 'uploading indenticon',
+      })
+
       const buffer = Buffer.from(await file.arrayBuffer())
-      console.log(buffer)
-      this.setState({ identicon: 'ipfs://' + (await ipfs.add(buffer)).path })
+      const picture_cid = await uploadFileToIPFSProxy({
```
xat, line 187: i'm not sure, but can't `file` just be passed directly to `uploadFileToIPFSProxy`?

src/pages/objkt-display/index.js
```diff
@@ -125,6 +127,36 @@ export const ObjktDisplay = () => {
   }, [])
 
   const Tab = TABS[tabIndex].component
+
+  const objkt_classes = []
```
xat, line 131: i think this needs to be put in `useState`.

xat (commented):
left a few comments, nothing dramatic :)

src/components/media-types/index.js
melMass: By default it's optional and using the ENV key to choose the gateway, the second argument can be used to specify the gateway per instance use

src/pages/objkt-display/index.js
melMass: Isn't making the useEffect depend on nft alleviates that?

melMass:
> left a few comments, nothing dramatic :)

Thank youu, making the edits now!

src/pages/config/index.js
melMass: Probably!

This was a bit of the trouble in this PR, that's why I roughly typed it `@typedef { {path: string?, blob: Blob} } FileHolder`, extracting it from the old code and it's usage mainly to make sense of the code. I would happily clean it further in a separate branch/PR later

src/pages/objkt-display/index.js
melMass: I've done all besides this as I'm not sure ;)

src/components/media-types/index.js
xat: ah, true, i missed that 👍

src/pages/objkt-display/index.js
xat: hmm, well, if it works, it works :)

as soon as you add something async to the effect, it will probably break though.

src/pages/config/index.js
xat: ok, yeah, no need that we need to block merging this pr because of that.

xat:
i manually tested the "Cannot edit profile" fix, it worked 👍 

if there is nothing left otherwise, i can merge.

melMass:
> i manually tested the "Cannot edit profile" fix, it worked 👍
> 
> if there is nothing left otherwise, i can merge.

Yep it solves a few important things like subfolders for generative and profile editing. Further edits can be made in another one ;)

xat (approved):
great work!

PR #163, opened 2022-08-19
fix: ⚡️ feeds

### LINKED ISSUE #110: [bug] Infinite scroll on "new OBJKTs" feed is broken

melMass:
This closes #110
Reasoning is simple, if 15 operation by the same creator or on the same token (depending on the feed) it would break the infinite scroll.
This PR:
- bumps the batch items count 40
- Don't uniq creators in `recent sales`, IMO a single creator can sell multiple tokens I don't see why we should filter that? Still doing uniq by ID to not have duplicate tokens.
- Completely stops uniq on `new OBJKT`

### DIFF
--- a/src/pages/search/index.js
+++ b/src/pages/search/index.js
@@ -196,7 +196,7 @@ async function fetchSales(offset) {
   const { errors, data } = await fetchGraphQL(
     `
   query sales {
-    trade(order_by: {timestamp: desc}, limit : ${FETCH_BATCH}, offset : ${offset}, where: {swap: {price: {_gte: "0"}}}) {
+    trade(order_by: {timestamp: desc}, limit : ${FETCH_BATCH}, offset : ${offset}) {
       timestamp
       swap {
         price
--- a/src/pages/search/index.js
+++ b/src/pages/search/index.js
@@ -345,8 +345,11 @@ export class Search extends Component {
   }
 
   update = async (e, reset) => {
-    const arr = getWalletBlockList()
+    const bans = getWalletBlockList()
 
+    const filterBan = (e) => {
+      return !bans.includes(e.creator_id)
+    }
     this.setState({ select: e })
     if (reset) {
       this.setState({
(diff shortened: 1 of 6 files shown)


src/pages/search/index.js
```diff
@@ -183,7 +185,7 @@ async function fetchSales(offset) {
   const { errors, data } = await fetchGraphQL(
     `
   query sales {
-    trade(order_by: {timestamp: desc}, limit : 15, offset : ${offset}, where: {swap: {price: {_gte: "0"}}}) {
+    trade(order_by: {timestamp: desc}, limit : ${FETCH_BATCH}, offset : ${offset}, where: {swap: {price: {_gte: "0"}}}) {
```
jagracar: This price filter can be removed, right? All prices are >= 0...

src/pages/search/index.js
```diff
@@ -446,23 +451,17 @@ export class Search extends Component {
...
-          _.uniqBy([...this.state.feed, ...tokens], 'id'),
-          'creator_id'
-        ),
+        feed: _.uniqBy([...this.state.feed, ...tokens], 'id'),
       })
     }
 
     if (e === 'new OBJKTs') {
       let tokens = await fetchFeed(this.state.lastId, this.state.offset)
-      tokens = tokens.filter((e) => !arr.includes(e.creator_id))
```
jagracar, line 475: I would actually add two conditions for the filter in the new OBJKTs feed.

We could have a dict to track how many tokens have been displayed in the feed for a given artist_address {artist_address: counter}. Then in each batch we do the following:
- if the artist counter is <= 10, then we display the token and increase the counter.
- if the artist counter is > 10, then we only display the first token of that artist in that batch and filter the rest.

In this way if a user mints 10 tokens in the same batch, the tokens will still be shown.

If a user mints 200 tokens, the first 10 tokens will be shown, and after that we will see only one token per batch (1 every 45). So, if one goes very down in the feed, we will still be able to find old mints for normal users, because we have the 1 per batch option.

This will also guarantee that we never get an empty batch. If the same artist minted the 45 tokens in that batch, only the first token will pass the filter.

PR #164, opened 2022-08-19
fix: ⚡️ Update links to COC

merchantcoppola:
We updated the COC especially the T&C part for the new NSFW/photosens moderation rules and in the process i had to change the subtitles, so new links are needed


melMass (approved):
📝 Thanks

PR #165, opened 2022-08-19
fix: ⚡️ root is a directory detection

melMass:
This allows user to provide a zip whose root is a folder, both options would be valid:


melMass:
Even for this case the root can be found:
image

PR #166, opened 2022-08-19
fix: ⚡️ metadata & moderation: continue

melMass:
- fix: 🔥 blurred tokens
  
- small style edit
  from:
  Screenshot 2022-08-19 at 22 35 53

  to:
  Screenshot 2022-08-19 at 22 35 32


melMass (commented):
Merging! We can use the same branch to add photosensitive filter once we have a bit of feedback on the matter.

PR #175, opened 2022-08-22
feat: 🎨 status banner

### LINKED ISSUE #152: [feat] Provide a way to display a status message on the marketplace web site
### Describe the problem

Currently, there isn't a way to display a status message on the marketplace other than deploying a new version of the UI.

### Describe the solution you'd like

Provide a text file hosted on GitHub that can be used to provide a status message. The marketplace UI will read this file and if its has any content, that will be displayed to users. (Just a simple suggested solution; other possible solutions should also be considered)

### Alternatives considered

_No response_

### Additional context

_No response_

melMass:
Fixes #152

It's reading the mardown from the https://github.com/teia-community/teia-status/blob/main/banner.md file, use '  ' (double space) at the end of lines to separate blocks.
The config file: https://github.com/teia-community/teia-status/blob/main/banner_config.json is used to set the color and if the banner should display or not

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "framer-motion": "^4.1.16",
         "ipfs-http-client": "^53.0.1",
         "ipfs-only-hash": "^4.0.0",
+        "json5": "^2.2.1",
         "local-storage": "^2.0.0",
         "lodash": "^4.17.21",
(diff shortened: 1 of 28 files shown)


xat (approved):
as discussed: approving and merging this one now.

PR #177, opened 2022-08-23
fix: ⚡️ add mintingTool metadata

LeonNicholls:
Requested by objkt.com to debug any minting issues.


melMass (approved):
That's actually a very good thing to have!!

PR #178, opened 2022-08-23
feat: ♿ first accessibility updates

LeonNicholls:
First updates for accessibility improvements based on the aXe extension: https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd

### DIFF
--- a/src/components/media-types/glb/index.js
+++ b/src/components/media-types/glb/index.js
@@ -7,6 +7,7 @@ export const GLBComponent = ({
   preview,
   onDetailView,
   displayView,
+  objkt,
 }) => {
   const ref = useRef()
   const [width, setWidth] = useState('100px')
@@ -45,7 +46,11 @@ export const GLBComponent = ({
   if (displayView) {
     return (
       <div className={styles.container} ref={ref}>
-        <model-viewer {...props} style={{ width, height }}>
+        <model-viewer
+          {...props}
+          style={{ width, height }}
+          title={`GLB object ${objkt}`}
+        >
           <button slot="ar-button" className={styles.arButton}>
             AR
           </button>
@@ -55,7 +60,7 @@ export const GLBComponent = ({
   } else {
     return (
       <div>
-        <model-viewer {...props}>
+        <model-viewer {...props} title={`GLB object ${objkt}`}>
           <button slot="ar-button" className={styles.arButton}>
             AR
           </button>
--- a/src/components/media-types/glb/index.js
+++ b/src/components/media-types/glb/index.js
@@ -7,7 +7,7 @@ export const GLBComponent = ({
   preview,
   onDetailView,
   displayView,
-  objkt,
+  objktID,
 }) => {
   const ref = useRef()
   const [width, setWidth] = useState('100px')
@@ -49,7 +49,7 @@ export const GLBComponent = ({
         <model-viewer
           {...props}
           style={{ width, height }}
-          title={`GLB object ${objkt}`}
+          title={`GLB object ${objktID}`}
         >
           <button slot="ar-button" className={styles.arButton}>
             AR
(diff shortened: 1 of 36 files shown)


src/components/media-types/glb/index.js
```diff
@@ -7,6 +7,7 @@ export const GLBComponent = ({
   preview,
   onDetailView,
   displayView,
+  objkt,
```
melMass: What does this refer to?

melMass:
I think there is a more "react" way to handle this! Very nice tool

https://reactjs.org/docs/accessibility.html

melMass:
This might help in spotting them directly in code:
https://reactjs.org/docs/accessibility.html#eslint-plugin-jsx-a11y

src/components/media-types/glb/index.js
LeonNicholls: It's the objkt ID. It's the same as with other existing components.

LeonNicholls:
Was going to add eslint-plugin-jsx-a11y in a later update, but I did it now.
Each tool that I use detects different issues with the UI. jsx-a11y seems happy with the changes recommended by the aXe extension, but raises various other possible design issues with the UI.

src/components/media-types/glb/index.js
melMass: Not sure to understand, can't we pass an object with more info?
I'll look into it when I have time, but again, to avoid having to add 10 new parameters to 10 files each time we want to pass something new to a function, not sure if it's clear I'll make a proposition

melMass:
> Was going to add eslint-plugin-jsx-a11y in a later update, but I did it now.
> 
> Each tool that I use detects different issues with the UI. jsx-a11y seems happy with the changes recommended by the aXe extension, but raises various other possible design issues with the UI.

Yeah that's expected.
I just feel like the hints from source code will be more react oriented than analyzing the built bundle that goes through transformations

src/components/media-types/glb/index.js
LeonNicholls: Well, that's the pattern followed by the existing components. Happy to pass the whole query result down in case other info might be needed in the future.

LeonNicholls:
> > Was going to add eslint-plugin-jsx-a11y in a later update, but I did it now.
> > Each tool that I use detects different issues with the UI. jsx-a11y seems happy with the changes recommended by the aXe extension, but raises various other possible design issues with the UI.
> 
> Yeah that's expected. I just feel like the hints from source code will be more react oriented than analyzing the built bundle that goes through transformations

Yeah, agreed, But jsx-a11y is not enough. The other tools raise issues with navigation, the minting form, and with visual issues that the plugin can't find.

melMass:
> jsx-a11y is not enough.

I'm no accessibility expert, but maybe we should prfioritize from what is the most needed for users? 
I'm asking ~~SurfingDarkness~~ SourOrchid if she has inputs on the most critical issues

melMass:
Linking here Synk's write up: https://www.notion.so/teia-community/Accessibility-UI-UX-87d2598ed49b49b785e904d0ec56ef61

LeonNicholls:
Thanks. That's why I started with the aXe extension since it verifies issues from a user's perspective.
It does look like even after using various tools, it is going to require feedback from users since it feels like I'm fixing technical issues and not yet all the usability issues.

src/components/media-types/glb/index.js
melMass: I would vote for that!
If not let's rename it to objktID to make it clearer

LeonNicholls: I looked at the code a bit. It seems that the queries that drive these components' data are not the same. They are also declared in various places. Maybe, for now, go with your suggestion to rename the variables to objktID, otherwise, I fear the changes, which have little benefit for now, will make any code refactoring unnecessarily complicated.

melMass: Yep, that's one of the big pains of the current codebase, lots of duplicated (slightly different, I guess by different devs) logic!

melMass (approved):
Nice, it revealed other things that could be greatly simplified (not related to your additions)

PR #188, opened 2022-09-06
fix: ⚡️ renamed galleries to _galleries

xat:


public/_galleries/NFXT_SPACE.json
```diff
@@ -1,37 +1,37 @@
...
-    },
-    {
-      "artist": "Nero One",
-      "wallet": "tz1UH9shZ3hSio2nuwpRPHwvockWa3XWMNA9",
-      "objkt": [4782, 24049]
-    }
-  ]
-}
+{
+  "title": "NFXT SPACE",
```
melMass, line 2: All these reformats are due to the fact that when we first ran the formatter over the whole codebase, we where limiting to js files only

melMass (approved):
Amazing!

PR #196, opened 2022-09-26
feat: 🇵🇰 pakistan feed

melMass:


src/pages/search/index.js
```diff
@@ -316,16 +316,17 @@ export class Search extends Component {
     flag: false,
     lastId: undefined,
     tags: [
-      { id: 11, value: '🇺🇦 ukraine', label: 'ukraine' },
-      { id: 12, value: '🏳️‍🌈 tezospride', label: 'tezos pride' },
+      { id: 10, value: 'pakistan', label: '🇵🇰 pakistan' },
```
LeonNicholls: These are used as aria-label values for accessibility. The emoticons don't work with the text-to-speech systems, so should be avoided. Just keep textual info for the labels.

melMass: We could either use the value or another key, but the label/value was used the other way around. This also avoids emoticons check in ifs

melMass: @NoRulesJustFeels fixed in the last commit

xat (approved):
don't have the time to properly review. but merch said, that he manually tested it. so merging it now.

PR #202, opened 2022-10-06
temp switch hicdex and and ipfs-proxy to teztok server so that zir0h …

xat:
…can update his server


xat:
i suppose we can also just risk doing this without huge testing sessions :)

Zir0h:
Or on one of those onrender previews?

Zir0h:
Nevermind, YOLO'd on .rocks, seems to work perfectly! :smile:

PR #207, opened 2022-10-19
feat: 🚀 add Tezos for Iran

melMass:
- Separated tag row from fundraise row.
- Replaced a tags with buttons (not component yet)


melMass:
image

xat (approved):

PR #218, opened 2022-11-27
fix: ⚡️ replace infinite-scroll lib

melMass:
very untested for now


src/pages/gallery-detail/index.js
```diff
@@ -118,11 +118,8 @@ export const GalleryDetail = () => {
           <Container xlarge>
             <Padding>
               <InfiniteScroll
-                dataLength={context.feed.length}
-                next={loadMore}
-                hasMore={hasMore}
-                loader={undefined}
-                endMessage={undefined}
+                loadMore={loadMore}
+                hasMore={offset < context.feed.length}
```
xat, line 120: hmm, infinite scroll doesn't seem to work here (wondering if it's even needed).

xat:
i'll merge it now. we con do further fixing in develop.

PR #244, opened 2023-02-01
fix: migrate to objktcom public API v3

vhf:
Hey, I noticed listings are somewhat broken following the planned shut down of objktcom v2 public API. Here's a fix using v3 instead.


xat (approved):
thx!

PR #248, opened 2023-02-08
ci: 🚀 docker image workflow

Zir0h:
Currently this only pushes to my private repo, but we could add docker hub later on.


melMass (approved):

melMass:
Great!
I'm going to merge, but I'm just not sure about relative paths, I believe they are from the root of the repo, not `.workflows` easy fix if needed

PR #256, opened 2023-02-16
fix: ⚡️ new minter

melMass:
Still Draft 😄 

Should be cleaned up / ready for preview later today.


melMass:
Closing, the same commits were pushed on `develop`

PR #262, opened 2023-02-21
🧪 exp: gql types

melMass:
> **Warning**
> Do not merge. This is a W.I.P

- [ ] convert to typescript (wip)
- [x] add API type generation on build (let's keep in in version control `src/bindings.ts`, thanks to husky it should only update when teztok changes something)
- [x] add eslint on push (we should maybe scope it to some branches)

--- 

*update from may 24*
- [ ] finish or revert (loosen) the typescript conversion. All recent commits where done using `--no-verify` 
- [ ] related to the first task, but I think the whole codegen thing was a bad idea after playing with it in teztok/teia's context. so same either finish or remove.
- [ ] clean leftovers from the UI update.


melMass:
Alright, I'll finish this this week!

melMass:
> Alright, I'll finish this this week!

melMass:
Goals:
- [ ] finish or revert (loosen) the typescript conversion. All recent commits where done using `--no-verify` 
- [ ] related to the first task, but I think the whole codegen thing was a bad idea after playing with it in teztok/teia's context. so same either finish or remove.
- [ ] clean leftovers from the UI update.

PR #265, opened 2023-03-01
fix: 🐛 nginx config username can also contain dots

Zir0h:
Right now the previews aren't being triggered for https://teia.art/Generative.Hut/collection
This commit fixes the regex to also include dots


melMass (approved):
Zir0h's nightmares contain regex 😂

Zir0h:
> Zir0h's nightmares contain regex joy

it's true :rofl:

PR #286, opened 2023-04-03
feat: ✨  bring back fool + matrix theme

melMass:


melMass:
Let's keep PR short as we said 👼 .
This is done

Zir0h (approved):

PR #290, opened 2023-04-11
docs: 📝 collab message for non managers

ktorn:
This original message:
> You aren’t part of any collaborations at the moment

was confusing, because the user may be a member of collabs, but not a manager of those contracts.

The new message makes it clear:
> You are not a manager of any collaborations at the moment


melMass (approved):
Thanks

PR #293, opened 2023-05-02
feat: Preconnect to lists.teia.art and imgproxy.teia.rocks

Zir0h:
I noticed only cache.teia.rocks and raw.githubusercontent.com were in the list. Every full pageload needs multiple requests to lists.teia.art, so it might be a good idea to also preconnect it?


Zir0h:
Also added imgproxy.teia.rocks

melMass (approved):
LGTM

PR #295, opened 2023-05-04
fix: use default ipfs context root for the CDN

Zir0h:
There's probably some more code cleanup needed from our experiments with the tcinfra thumbnail generator.


melMass:
There should not no or very little! but if I remember right we don't really use this anymore directly

melMass (approved):

PR #297, opened 2023-05-11
filter out burned tokens in the new-objkts feed

xat:


Zir0h (approved):
looks good to me!

melMass (approved):
I was just not sure about the added line in swap but it renders well!
image

PR #305, opened 2023-06-15
Fix: switch to lists.teia.art

Zir0h:
Because github is blocked in India? I don't know... But this would remove a dependency that we don't control.


melMass (approved):
🦸

PR #308, opened 2023-07-04
Fix: remove dead rpc nodes, added teia node

Zir0h:
We got a report from a user that was unable to mint a piece in discord, they were using a broken rpc node in the settings. Switching to another node fixed it.


xat (approved):

melMass:
So this removes custom?

LeonNicholls:
I vote for keeping custom.

Zir0h:
I'm sorry, I thought that was a placeholder. How can I use the custom option? :/

PR #310, opened 2023-07-04
Fix: add back the custom rpc node setting

Zir0h:
I deleted this the last rpc node update, I thought this was a placeholder. I was never able to use a custom rpc node without adapting it myself in localstorage. I probably missed something obvious :pray:


melMass:
The idea was to show an extra text input when custom is selected but it doesn't seem to be in either main or develop so it's probably something I only did locally... I can't recall.
I'll add it tonight or tomorrow to this PR

melMass (approved):
I'll merge in a few minutes 💌

PR #312, opened 2023-07-29
Adds the initial structure for the teia dao token claim page

jagracar:


jagracar:
Hi Mel an others. I had to comment the pre-commit code to be able to commit my changes...

This is just the structure to be filled for the token claim page. Feel free to edit and improve it. The utils.js file contains the code (commented) to do the claim button work. It needs some context info, so it probably needs to be put somewhere else

jagracar:
I uncommented the pre-commit part using the github UI and it worked XD

melMass:
Small tip you can use "--no-verify" to bypass hooks, if you can post the log of why it did not allow it it could be fixed I think!

src/pages/dao/utils.js
```diff
@@ -0,0 +1,95 @@
...
+    0: "QmR3tiFDi8vgTdvR98jS8jLn2No8EheMG6WTnpbzbupvZm",
+    1: "QmWG2jqiHXFk6Ub7xY7TtjYomxern3SNXuCDVxVeFc88iA",
+    2: "QmZ6P44rhTMxKHqADcW27quQPgvs6h1YTQaz1hS8J3FggF",
+    3: "Qmbm99hPZBQKg5wUhzYTuQUrit5eDgvytqU8J17kpcC71T",
+    4: "QmVBQjr7m25MgF2S7BkmNMtzVNfSuWCaxsYJMXkA7PNhKu",
+    5: "QmTi6E5HJJGZrwKkvBRYMHx2DaVD9mij9rm5cSyDVWv64f",
+    6: "QmbZsvCDiGv4xWeLtMf1wPxjogpvLjaKLwdZe5xQezUWUS",
+    7: "QmbCvvyP2s6nqq1cCW9H617jRGEYoZkV3SoJHCS9sJmsNU",
+    8: "QmRhr7m6zBtU1YFNcLMo34jQVUFdamG3BCGCFjA9AwBEnK"
+};
```
Zir0h: I pinned these so that they can be fetched with cache.teia.rocks

melMass:
Very clean!! 👏 
What is the next step @jagracar ?

src/pages/dao/utils.js
jagracar: Those are from the test we did one month ago. We will need to create new ones with the final distribution

jagracar:
We need to put the js code in utils.js somewhere else where it can access the context. Also add all the text explanation about the token claim and it's implications (merch has something already).

src/components/header/main_menu/MainMenu.jsx
```diff
@@ -63,6 +63,13 @@ export const MainMenu = () => {
             route="subjkt"
             need_sync
           />
+
+          <MenuItem
+            className={styles.menu_label}
+            label="Claim DAO tokens"
+            route="dao"
+            need_sync={!currentName || !currentAddress}
```
jagracar: Not sure if this line is correct. What do you think @melMass?

src/context/userStore.ts
```diff
@@ -403,6 +409,30 @@ export const useUserStore = create<UserState>()(
...
+        claimTokens: async () => {
+          const user_address = get().address
+          const handleOp = get().handleOp
+          const showError = useModalStore.getState().showError
+          const step = useModalStore.getState().step
+
+          step('Claiming Teia DAO tokens', 'Claiming Teia DAO tokens', true)
+
+          // Download the distribution file mapping file
+          const mapping = await downloadFileFromIpfs(DISTRIBUTION_MAPPING_IPFS_PATH)
```
jagracar: Doesn't handle possible download errors...

src/context/userStore.ts
```diff
@@ -403,6 +409,30 @@ export const useUserStore = create<UserState>()(
...
+
+          // Download the Merkle data
+          const merkleDataPath = MERKLE_DATA_IPFS_PATHS[mapping[user_address]]
```
jagracar: Doesn't handle the possibility that the user_address is not in the mapping dictionary. That could happen with new users

src/context/userStore.ts
```diff
@@ -403,6 +409,30 @@ export const useUserStore = create<UserState>()(
...
+          const merkleData = await downloadFileFromIpfs(merkleDataPath)
+          console.log(merkleData)
+          try {
+            const dropContract = await Tezos.wallet.at(DAO_TOKEN_CLAIM_CONTRACT)
+            const batch = dropContract.methods.claim(merkleData.proof, merkleData.leafDataPacked)
```
jagracar: This fails at the moment because the current claim contract was deployed to the testnet and not to the mainnet

src/pages/dao/index.jsx
```diff
@@ -0,0 +1,30 @@
...
+  const claimTokens = useUserStore((st) => st.claimTokens)
+
+  return (
+    <Page title="DAO" large>
+      <div className={styles.dao__outer__container}>
+        <h1 className={styles.dao__title}>TEIA DAO</h1>
+      </div>
+
+      <div className={styles.dao__outer__container}>
+        <p>Warning, warning, warning!</p>
```
jagracar: We should fill this with all the claim info. @merchantcoppola has written already something

jagracar:
Hi @melMass and @xat This already should be working once the token claim contract is deployed to the mainnet. Could you check that i'm not doing something stupid? I put the code where i thought it fit best, but i'm not totally sure.
There are still many things to improve, but the general structure should be there. Feel free to change things.

src/context/userStore.ts
jagracar: Fixed that in my last commits

src/context/userStore.ts
jagracar: Fixed that in my last commit

jagracar:
This is how the claim page looks at the moment

claim

One option would be to remove/comment the direct link to the page in the menu and merge it. We will be able to test the page on production under teia.art/dao, but normal users will not see it

melMass:
Nice I will be able to look into it tomorrow if still needed!

jagracar:
> Nice I will be able to look into it tomorrow if still needed!

Yes, please!

melMass:
Will review today :) 
From a quick look it looks all nice but I would just use markdown for the body like we do for the about: https://github.com/teia-community/teia-ui/blob/main/src/pages/about/index.jsx which is sourcing this md: https://github.com/teia-community/teia-ui/blob/main/src/lang/en/about.md

jagracar:
> Will review today :) From a quick look it looks all nice but I would just use markdown for the body like we do for the about: https://github.com/teia-community/teia-ui/blob/main/src/pages/about/index.jsx which is sourcing this md: https://github.com/teia-community/teia-ui/blob/main/src/lang/en/about.md

Oh, i used the FAQ as reference :(

I guess we can put the introduction and the legal disclaimer inside 2 md files. But the page has some interactivity, via the checkbox.

melMass:
> I guess we can put the introduction and the legal disclaimer inside 2 md files. But the page has some interactivity, via the checkbox.

Yep agreed if there is a need for using markup in places splitting it makes sense. Is there anything else besides the claim button? I'm not sure the staging link is getting auto update so I'm seing only that for now but saw the checkbox mention :)

src/utils/ipfs.js
melMass, line 1: @Zir0h would it make sense to use something else than ipfs.io?

@jagracar I think we can use this method to avoid hardcoding it: https://github.com/teia-community/teia-ui/blob/e18769a602c9210af12859e17db14bb8152dbf90/src/utils/index.ts#L40-L70

src/pages/dao/index.jsx
melMass, line 1: @jagracar follow up on the PR comments, if we go the md route it seems like interactivity is only at the very end so no need to split imo

melMass (changes requested):
Looks very good to me! Awesome work everyone.
My only "complaint" is using markdown but besides that it all looks good 🥇

jagracar:
> > I guess we can put the introduction and the legal disclaimer inside 2 md files. But the page has some interactivity, via the checkbox.
> 
> Yep agreed if there is a need for using markup in places splitting it makes sense. Is there anything else besides the claim button? I'm not sure the staging link is getting auto update so I'm seing only that for now but saw the checkbox mention :)

I will not touch the code for the next hours to give you time for the review. I will try to add the md files after that.

src/pages/dao/index.jsx
jagracar: Yep. Before there was a button in the middle, but not it's all at the end

jagracar: Oh, you already added the md files! Thanks!

src/utils/ipfs.js
jagracar: Not sure to which file do you refer to? The code in utils/ipfs.js ? I realized that that method is not very good. It assumes that the file that it downloads from ipfs is a json file....
it probably should be renamed to downloadJsonFileFromIpfs

I also don't know if there is a better place for that method.

melMass: In `src/utils/index.js` you have a `CIDtoURL` method that will use the env `VITE_IPFS_DEFAULT_GATEWAY` to resolve a gateway. this is just to avoid hard coding it but not a big deal at all

Zir0h: yeah I would prefer using `VITE_IPFS_DEFAULT_GATEWAY`

jagracar: Done in my last commit

melMass: I will be able to dedicate a bit of time tomorrow after lunch!

melMass (approved):

melMass:
@merchantcoppola you can keep doing the fixes for text in this branch, I'll update it in #313 since we have a live preview here: https://teia-community.github.io/teia-ui/previews/pr-313/ and I will push the extra changes from #313 here before merging

melMass:
@jagracar can you change your remote to here: https://github.com/teia-community/teia-ui/tree/feature/add-token-claim-page ? (see #318)

melMass:
@jagracar Can you check that the hooks are reworking now? I fixed them since, because it makes it quite hard to diff stuff

https://github.com/teia-community/teia-ui/pull/312/commits/bea45a5e7e5108a43e113e8e7cdfec87bd0630d5

jagracar:
How do i check that? I merge what you have in master in teia-ui? And then try to build it? I

melMass:
> How do i check that? I merge what you have in master in teia-ui? And then try to build it? I

You can change any arbitrary js ts jsx file, stage it and commit it, the hooks should run or fail, you can just uncommit the test afterward

melMass:
Just closing this one as it's getting a bit hard to follow

jagracar:
Tested it with this commit:
https://github.com/teia-community/teia-ui/pull/318/commits/03eaf003b81a2b9f772b4e457ac1a7245916c96f

PR #316, opened 2023-08-08
ci 🧹:  Scope & permissions checks

melMass:
- [x] Scope docker build to src files since we now added `worflow_dispatch`
- [x] Check permissions, there are actions on the marketplace but [this](https://michaelheap.com/github-actions-check-permission/) looks better


melMass:
Merging this to be able to test the actions it has no impact on critical stuff

PR #318, opened 2023-08-09
[feature] Adds the initial structure for the teia dao token claim page

melMass:
Migrated from #312


src/pages/dao/index.jsx
```diff
@@ -113,16 +113,16 @@ export const DAO = () => {
...
               distribute profits among its members. Teia does not sell TEIA
               tokens, nor to generate profit with the tokens. Claiming TEIA
               tokens via this claim page is free.
             </li>
 
             <li counter="2">
               <b>No investment advice:</b> The information provided on the claim
-              page, including but not limited to any articles, blog posts on
-              blog.teia.art, or other related materials, is for informational
+              page bor any articles, blog posts on blog.teia.art, or other related materials
```
melMass: Typo here @merchantcoppola but not sure what it should be, maybe "for" but it would be strange in context?

merchantcoppola: thx, i did a small update for that, do you think its better now?

melMass: Yes thanks!

PR #323, opened 2023-08-21
adding a troubleshooting section

merchantcoppola:
i notoced a lot of similar issues mentioned on discord, maybe we should give the most common solutions directly on the page?

PR #324, opened 2023-08-21
Increases the default storage fee to 400

jagracar:
Some claim operations fail because the storage fees set by the wallets are too low. I added a default storage fee to the claim operation to try to fix it.


melMass (approved):
Great, I'll merge #323 before this one

Issue #301, opened 2023-05-31
[feat] Home feed configuration


melMass:
### Describe the problem

I want to choose the first feed I'm welcomed with on teia.art

### Describe the solution you'd like

Probably just added there as a local setting for now: 
https://teia.art/settings

### Alternatives considered

Whatever we might choose for a better option (IPNS was mentioned, "hijacking" the subjkt json etc...)

### Additional context

🫶

caaatisgood:
hey @melMass could you help me understand the options of the first feed we're looking to let users choose? would love to help on this improvement!

melMass:
Hey @caaatisgood thanks a lot for the proposal, that would be great.
It shouldn't be too hard. I wanted to wait my [huge PR](https://github.com/teia-community/teia-ui/pull/262) before addressing these but I don't think it makes much sense now and I'll have time soonish to address it properly.

For now let's go with option 1: use the browser local storage! Here are the rough steps, don't hesitate to ask questions:

1. We have this store here that uses `Zustand` and is automatically tied to local storage, so it should be a matter of adding the proper key for the type and the setter for the store (for instance `startFeed` or something like that):
https://github.com/teia-community/teia-ui/blob/e18769a602c9210af12859e17db14bb8152dbf90/src/context/localSettingsStore.ts#L68-L157 

2. Then to actually add an UI for it you can add a `Select` component in the `Settings.jsx` file (check the rpc line to see how to use it):
https://github.com/teia-community/teia-ui/blob/main/src/pages/config/Settings.jsx

3. Finally to actually use the setting's value as the actual start feed, you will need to edit this line:
https://github.com/teia-community/teia-ui/blob/e18769a602c9210af12859e17db14bb8152dbf90/src/index.jsx#L79 with the store value (you will probably need a map between `string` -> `feedComponent`, a bit like [this](https://github.com/teia-community/teia-ui/blob/e18769a602c9210af12859e17db14bb8152dbf90/src/components/header/feed_toolbar/FeedToolbar.jsx#L26-L48) but with the names as key and the component as value

caaatisgood:
Hey @melMass thank you so much for the walkthrough. Very very useful! Just wanted to make sure that I understand the whole user flow and edge cases here.

## What we want to achieve

[...]

PR #325, opened 2023-08-23
feat: home feed configuration

caaatisgood:
https://github.com/teia-community/teia-ui/issues/301

Make the home feed customizable via `/settings`. See demo vid here https://github.com/teia-community/teia-ui/issues/301#issuecomment-1670335023.

I ended up importing all the feed comps in `home/index` to create a component map. Not the best approach performance-wise but does the job. Happy to do any CRs. (:


melMass:
Thanks! I should be able to review it before next week if no one can earlier, by a quicklook it seems nice, I would just remove the feed names map that is not really needed unless I got the logic wrong

src/constants.ts
```diff
@@ -68,6 +68,27 @@ export const ALLOWED_FILETYPES_LABEL = Object.entries(MIMETYPE)
   .map((e) => (e[0] === 'ZIP' ? 'HTML (ZIP ARCHIVE)' : e[0]))
   .join(', ')
 
+export const FEED_MAP = {
```
melMass: I was referring to this, I don't think it's needed

caaatisgood: This is like a single source of truth shared between the settings page and the home feed component. And also to derive the allowed feed types. I don't know if there's a way to avoid having such a map.

melMass: I simplified it to just a list, I can't fully test it but I just pushed a few edits, I'll be able to properly test it tomorrow and merge then! Thanks again

caaatisgood: Thanks @melMass!

melMass (approved):

PR #329, opened 2023-09-11
Update the tez-quakeaid description to include Morocco

Zir0h:


melMass:
I did not follow sorry quite busy lately! The split is already setup?

melMass (approved):
🖤

merchantcoppola:
imo we can merge this it updates the address to the dropdown split menue for new collab contracts and adds morroco as a cause to the old tezquakaid pieces (which is correct since new funds from the old pieces will be donated to morroco

but we will need a new, seprate feed for the morroco specific pieces with the new address: 

KT1RwXEP8Sj1UQDHPG4oEjRohBdzG2R7FCpA

PR #332, opened 2023-09-24
Feature/show dao token balance

jagracar:
Hi! This PR adds the TEIA balance info in the user profile. I hope i followed the standard way to do it!


src/data/api.ts
```diff
@@ -322,6 +322,20 @@ export const GetUserMetadata = async (walletAddr: string) => {
...
+  const parameters = {
+    'token.contract': DAO_TOKEN_CONTRACT,
+    'token.tokenId': '0',
+    'account': walletAddr,
+    "select": 'balance'}
+  const response = await axios.get(`https://api.tzkt.io/v1/tokens/balances`, { params: parameters })
+    .catch(error => console.log('Error while querying the account token balance:', error))
+
+  return response?.data[0]? parseInt(response.data[0]) / 1e6 : 0
+}
```
jagracar, line 349: We probably can also use teztok here? I guess it doesn't matter that much

melMass: Nope it's actually not indexed and IIRC @xat mentionned it not being very adequate to store there so this is totally fine

melMass (changes requested):
Looks perfectly fine to me! I would just handle error on awaiting for the Dao token so that it doesn't bork everything  if tzkt is down and either display an error dialog or just default to -1, 0, X or ERROR for the count value displayed (or even hide it if missing simply)

src/pages/profile/index.jsx
```diff
@@ -55,6 +55,9 @@ async function fetchUserInfo(addressOrSubjkt, type = 'user_address') {
     user.subjkt = holder.name
   }
 
+  const daoTokenBalance = await getDaoTokenBalance(user.address)
```
melMass, line 58: I mean here by wrapping this in a try block

src/data/api.ts
```diff
@@ -322,6 +322,20 @@ export const GetUserMetadata = async (walletAddr: string) => {
...
+ * Get User DAO token balance
+ */
+export async function getDaoTokenBalance(walletAddr: string) {
+  const parameters = {
+    'token.contract': DAO_TOKEN_CONTRACT,
+    'token.tokenId': '0',
+    'account': walletAddr,
+    "select": 'balance'}
+  const response = await axios.get(`https://api.tzkt.io/v1/tokens/balances`, { params: parameters })
+    .catch(error => console.log('Error while querying the account token balance:', error))
```
jagracar: Is this not doing what you want @melMass ?

jagracar: It will return 0 tokens, unless i did something wrong.

Zir0h: Can the tzkt url be changed to `VITE_TZKT_API` ? If we ever need to switch tzkt instances, then we only need to change it in https://github.com/teia-community/teia-ui/blob/main/.env

melMass: Yes indeed, I did not get that the last `?` here is a conditional and not a null property fallback https://github.com/teia-community/teia-ui/blob/364e3c22c79642cecde127e9127f609f726b2045/src/data/api.ts#L337

Are you sure the hooks are running 🤣 ?

melMass (approved):
Mergeable for me, I would just double check hooks are ran I find it stange that it did not add a space here before the last `?`: https://github.com/teia-community/teia-ui/blob/364e3c22c79642cecde127e9127f609f726b2045/src/data/api.ts#L337

jagracar:
Yes. I was wondering about the hooks... I didn't have any error when committing, so i assumed they run fine...

Maybe is because i commit directly without staging when i use eclipse?

src/data/api.ts
jagracar: > Can the tzkt url be changed to `VITE_TZKT_API` ? If we ever need to switch tzkt instances, then we only need to change it in https://github.com/teia-community/teia-ui/blob/main/.env

Did it in my last commit

Zir0h: Awesome!

jagracar:
Not sure if i'm allowed to merge... I did another change and it seems that the hooks are disabled in my commits? I have no clue how to fix that...

melMass:
> Not sure if i'm allowed to merge... I did another change and it seems that the hooks are disabled in my commits? I have no clue how to fix that...

What is your OS? The hooks are installed on first `npm install` if you are using the same local folder since a long time just clone a fresh instance and it should work (I use both windows and mac)

jagracar:
> > Not sure if i'm allowed to merge... I did another change and it seems that the hooks are disabled in my commits? I have no clue how to fix that...
> 
> What is your OS? The hooks are installed on first `npm install` if you are using the same local folder since a long time just clone a fresh instance and it should work (I use both windows and mac)

I use linux... I installed the repo with npm install and npm start
some how my commits overwrite the hooks...

melMass:
> I use linux... I installed the repo with npm install and npm start some how my commits overwrite the hooks...

You can always run prettier manually but we should find why it's doing that, I personally commit from the command line, but GUI tools shouldn't bypass it unless specifically asked to, maybe you disabled it when you initially had the issue and the config was kept or something like that

jagracar:
How do you run it locally?

npm lint ?

jagracar:
I cloned again the repo to be sure before my last commit and it still didn't run the hooks.

melMass:
> How do you run it locally?
> 
> npm lint ?

`npm run format`

jagracar:
Ok. That did work. I will use it. Even from a clean install i got 2 files that needed format changes: 
* src/styles/reset.scss
* src/components/header/sample_events.js

PR #337, opened 2023-10-18
Switch the default rpc node to mainnet.teia.rocks

Zir0h:
mainnet.api.tez.ie has been down for a couple of hours already, people are going to start to notice pretty soon :sweat_smile:

### DIFF
--- a/.env
+++ b/.env
@@ -16,4 +16,4 @@ VITE_IMGPROXY='https://imgproxy.teia.rocks'
 VITE_IPFS_UPLOAD_PROXY='https://ipfsproxy.teia.rocks'
 VITE_IPFS_DEFAULT_GATEWAY='CDN'
 
-VITE_TEZOS_RPC='https://mainnet.api.tez.ie'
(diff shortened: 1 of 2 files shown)


melMass:
Tchoop tchoop! 
What is the reasoning for moving the keys around? If it's to "force" reset user config we can just use versioning on the config

Zir0h:
> Tchoop tchoop! What is the reasoning for moving the keys around? If it's to "force" reset user config we can just use versioning on the config

No I thought the order was the same as the array, I didn't find the `  rpcNode: rpc_nodes[0],` until after the third try :sweat_smile:

melMass:
My bad I looked at the commits not the final diff so I was confused.

melMass (approved):
Rocks!

Zir0h:
> My bad I looked at the commits not the final diff so I was confused.

yeah sorry, I was fumbling around a bit :laughing: 
anyway, it looks like the original node is fixed now https://mainnet.api.tez.ie/chains/main/blocks/head/header but I think switching would still be a good idea
that node hasn't been very stable

PR #338, opened 2023-10-19
fix: some people have username.js as subjkt, this overrides the content-type

Zir0h:
@jagracar found this profile which is currently broken because of the .js suffix :sweat_smile: https://teia.art/marcmaurer.js
2023-10-19-165210_676x416_scrot

### DIFF
--- a/.github/workflows/nginx.conf
+++ b/.github/workflows/nginx.conf
@@ -92,10 +92,10 @@ server {
 
     location ~ "^/([0-9a-zA-Z\-_\ \.]+)" {
         default_type  text/html;
-        types {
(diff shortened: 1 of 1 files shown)


Zir0h:
not testable on the GitHub pr pages I'm afraid, it's only when the visitor directly accesses the profile

melMass:
> not testable on the GitHub pr pages I'm afraid, it's only when the visitor directly accesses the profile

Just after the review of Jag's PR I want to solve this issue, it should be doable with a vite config change based on the base url

melMass (approved):
🚀

PR #339, opened 2023-10-20
feat: POC loadbalancer with healthcheck for a couple of rpc nodes

Zir0h:
This exposes https://teia.art/rpc/ as an rpc endpoint with the following nodes behind it:

https://mainnet.teia.rocks/
https://rpc.tzkt.io/mainnet/
https://mainnet.tezos.marigold.dev/
https://mainnet.api.tez.ie/

An active healthcheck will see if the nodes are healthy every 10s by probing the `/rpc/chains/main/is_bootstrapped` page.


melMass:
This is a bit chinese to me 😅 ! But looks cool 👀 
Should we set it as the new default? I can add it simply to the config versioning callback so everyone will be set to it

Zir0h:
> This is a bit chinese to me 😅 ! But looks cool 👀 Should we set it as the new default? I can add it simply to the config versioning callback so everyone will be set to it

I'd rather let it run for a couple of days and see how it behaves first :sweat_smile:

LeonNicholls:
Yes, agreed on testing it for a while first. It won't be long before one of the RPC nodes has issues, so it would be good to confirm that the logic works as expected when that happens.

melMass (approved):

Zir0h:
A couple of days turned into a couple of months, I've been using the teia.art/rpc LB as a daily driver since it was created (both on teia.art and as my wallet rpc node). I've had no issues, but I did see some slower rpc calls. I've put both teia nodes as primary and the rest as fallback in the meantime and that fixed the slower calls.

Issue #280, opened 2023-04-01
[bug] out of memory crash when uploading a large mp4 on chrome


Zir0h:
### Describe the bug

When minting a large mp4, you're able to select the file, but after a couple of seconds the page crashes.
Smaller mp4's are working fine, the "upload cover button" never appears

### Reproduction

1 sync + go to the mint page
2 select a >1GB mp4 file
3 wait around 15seconds

### Expected behavior

The page should not crash

### Platform and versions

```shell
Chromium 111.0.5563.146 Arch Linux
```

### Console output

```shell
n/a
```

melMass:
Thanks for logging it, yep it's a long lasting planned 'todo', for now I pass the whole file as bytes to the preview components 😅 .... This is pasted raw in the DOM and crashes.
The solution is to create an [object url](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL), it should be trivial:

AlexandreRangel:
Bump : )
Hi Teia friends! Any news on this issue?

melMass:
I will take a look in the weekend!

AlexandreRangel:
Hi! Is the 1GB upload limit still on the plans?

melMass:
> Hi! Is the 1GB upload limit still on the plans?

The limit is now more than that but the mechanism we use for previewing vidéo before mint creates OOM errors.
There are better approaches to do it and it's planned but no ETA yet, the issue is already elevated to high priority

AlexandreRangel:
Great!! What's the current limit?

Zir0h:
> Great!! What's the current limit?

2GB, been that way for a while now. It's only problematic if you run into this out of memory bug :sweat_smile:

PR #346, opened 2023-10-27
fix: avoid out of memory when minting large files

Zir0h:
I tried to fix the out of memory bug from #280 with the suggestion @melMass posted. I don't know if I implemented it correctly tough. It seems to work as expected.

### DIFF
--- a/src/components/form/MintForm.jsx
+++ b/src/components/form/MintForm.jsx
@@ -29,6 +29,8 @@ export default function MintForm() {
           .show(`File too big: ${data.artifact.file.size / 1e6}mb`)
         return
       }
+      const URL = window.URL || window.webkitURL
+      data.artifact.reader = URL.createObjectURL(data.artifact.file)
     }
     useMintStore.setState({ ...data, isValid: true })
     navigate('preview')
--- a/src/pages/mint/fields.js
+++ b/src/pages/mint/fields.js
@@ -141,7 +141,7 @@ export const fields = [
     type: 'file',
     watch: true,
     rules: {
-      required: 'No file selected',
+      required: 'You did not select a valid file',
     },
   },
   {


melMass:
You are on fire ❤️!
I just missed my train so I'll have a look a bit later today

melMass:
I'm getting 

`Uncaught (in promise) DOMException: The requested file could not be read, typically due to permission problems that have occurred after a reference to a file was acquired.`

For a 5Gb file.

As I'm looking into it it seems that chrome for instance has an ArrayBuffer size limit of 2GB... So we need to chunk it somehow 😓

Zir0h:
> I'm getting
> 
> `Uncaught (in promise) DOMException: The requested file could not be read, typically due to permission problems that have occurred after a reference to a file was acquired.`
> 
> For a 5Gb file.
> 
> As I'm looking into it it seems that chrome for instance has an ArrayBuffer size limit of 2GB... So we need to chunk it somehow 😓

Good thing we limited minting to 2GB :rofl:

melMass:
Oops! But with a 1.5GB I get this without errors in the console though

image

melMass:
Ohhh ok there was just a delay between the time it catched the first one to then display the cover... For large file there should probably be some kind of loader to indicate that

melMass:
In any case not related to this PR, it works!
But we should improve the mint UX a bit imo

Zir0h:
> But we should improve the mint UX a bit imo

I agree 100%! But I'm not good with frontend stuff :laughing:

melMass (approved):
Perfect!

In a next PR we should probably:
- check the size of the file and alert about that
- Add some kind of load state after selecting the main mint file, as the UI Isn't blocked one can click submit which fails the validation, not a big deal since once loaded the button for the cover appears and the mint state is restored

PR #347, opened 2023-10-27
moved some files around for the docker build

Zir0h:
The emails about the failed workflow "tzproxy.yaml" are getting a bit annoying :sweat_smile:


melMass (approved):
Ok we will accept the 🐳 in root 💘

Issue #311, opened 2023-07-11
[feat] confusing UI for the transfer tab


Zir0h:
### Describe the problem

There have been a couple of people having trouble sending tokens with the `tranfer` tab. 

When you add an address to the list, you need to click the `Add` button, otherwise it doesn't get "commited" to the list.
Because of this, it can happen that the user forgets to click on the `Add` button for the final recipient. In this case the last address will be ignored.

### Describe the solution you'd like

Maybe replace the add button with an `onfocusout` event?

### Alternatives considered

Replace the `Add` text with something like "Save input to list" / "Commit to list"? I'm not good with frontend stuff :sweat_smile:

Zir0h:
In this screenshot for example, the last address will be ignored if I hit "Send"
![2023-07-11-173413_1836x693_scrot](https://github.com/teia-community/teia-ui/assets/546070/b649335f-8280-4594-83b7-99d7e9cd4168)

melMass:
There could be a way by checking that the address is valid on loose focus! The count in the header was meant to show that but I agree!

Zir0h:
> There could be a way by checking that the address is valid on loose focus! The count in the header was meant to show that but I agree!

or maybe check the validity of the address each keystroke (with debounce ofcourse) and change the field color or something like that? just throwing some stuff out there :laughing:

Zir0h:
Fixed with https://github.com/teia-community/teia-ui/pull/348

PR #348, opened 2023-10-27
fix: add address to the list when the visitor clicks away from the input field

Zir0h:
possible fix for https://github.com/teia-community/teia-ui/issues/311


melMass (approved):
Lgtm!

Issue #336, opened 2023-10-17
subjkt settings [bug] 


Synkitten:
### Describe the bug

When adding your subjkt details for a new wallet I found that the first letter of the chosen name is removed automatically if you use a capital e.g. 'Moments' is automatically changed to 'oments'. The biggest problem is it takes a few seconds for the auto recognition and removal of the capitalised letter which isn't obvious if you have moved on to filling out the next questions. 
If you check at the wrong time you don't realise the issue before accepting the transaction.
![name issue 2](https://github.com/teia-community/teia-ui/assets/94412910/c04cc8f9-9087-447f-bcc5-71444ee7ab8f)

### Reproduction

1. Type name in with a capital
2. Wait for letter to disappear

### Expected behavior

Make capitals useable if it's possible but otherwise make it autocorrect to lower case or have a notice show if it is deleted so you know to check maybe.

### Platform and versions

```shell
OS windows
Browser - Brave (chrome)
Version - v. 1.58.137
```

melMass:
@Zir0h @jagracar I will check properly but like that I can't recall what we ended up with, kind of related to #334 : Are Subjkt case sensitive?

Zir0h:
This is actually a duplicate of https://github.com/teia-community/teia-ui/issues/303
The UI silently drops "invalid" characters, currently only lowecase and 0-9 is allowed I think.
Most important part is that the visitor should receive feedback about errors, rather than the site silently dropping bad characters.

melMass:
Thanks but my question was about the Subjkt casing at the contract level

jagracar:
> Thanks but my question was about the Subjkt casing at the contract level

I don't think so. The entry point takes bytes and i don't think it does any string manipulation. My guess is that is possible to register jagracar, Jagracar or JAGRACAR using better call dev.
https://tzkt.io/KT1My1wDZHDGweCrJnQJi3wcFaS67iksirvj/entrypoints

In fact, you can see some capitals here:
https://tzkt.io/KT1My1wDZHDGweCrJnQJi3wcFaS67iksirvj/storage/3920

I even find some recent ones with capitals. Maybe they were registered in one of the h=n forks.
https://better-call.dev/mainnet/KT1My1wDZHDGweCrJnQJi3wcFaS67iksirvj/storage/big_map/3920/keys

melMass:
> My guess is that is possible to register jagracar, Jagracar or JAGRACAR using better call dev. 

[...]

Issue #303, opened 2023-06-07
[bug] when changing your subjkt username, the field silently drops invalid characters


Zir0h:
### Describe the bug

When changing your subjkt username, the field silently drops invalid characters. This causes confusion with the end users. I think it might be better to block the form submit and show a warning. In that case the user can fix the username themself.

### Reproduction

1. go to https://teia.art/subjkt
2. insert a username what contains invalid characters (capital letters for example)
3. the invalid characters get dropped silently when unfocussing the field

### Expected behavior

The user gets a warning message with directions on how to fix their username

### Platform and versions

```shell
- Linux
- Firefox 113.0.2 (64-bit)
```

melMass:
The kind of bug I hate as a user! I will look into it later today

PR #349, opened 2023-10-27
Allow capitalization and show message when the subjkt contains invalid characters

Zir0h:
Possible fix for https://github.com/teia-community/teia-ui/issues/303 and https://github.com/teia-community/teia-ui/issues/336

### DIFF
--- a/src/pages/config/Subjkt.tsx
+++ b/src/pages/config/Subjkt.tsx
@@ -266,11 +266,13 @@ export const Subjkt = () => {
               value={subjkt}
               onChange={setSubjkt}
               onBlur={(e) => {
-                setSubjkt(e.target.value.replace(/[^a-z0-9-._]/g, ''))
+                if(/[^A-Za-z0-9-._]/.test(e.target.value)) {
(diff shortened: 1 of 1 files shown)


melMass:
Awesome thanks!
So if I get this right we allow uppercase finally? 
I'll have to check more closely but the regex test that spawns the dialog seems wrong no? The `^` should be outside the brackets I think

Edit: maybe not at least it was like that before minus uppercase

Zir0h:
> Awesome thanks! So if I get this right we allow uppercase finally? I'll have to check more closely but the regex test that spawns the dialog seems wrong no? The `^` should be outside the brackets I think

Yeah, I added uppercase again.

If you move the ^ outside of the brackets you change its meaning to "starts with". Inside the brackets it's "does not contain".
In this case the code tests if it finds characters that are not in the allowed list. 

So it's the inverse of the placeholder to show the error. I suck at explaining :rofl:

melMass (approved):
Zir0h? More like RegeXoh! Thanks

Issue #278, opened 2023-03-29
[bug] GIFs uploaded twice


merchantcoppola:
### Describe the problem

Current behaviour: If you mint a large gif, it will duplicate the same gif for the thumbnail by automatically uploading it again.
Concerns are that folks wont wait for huge thumbnail to load.

### Describe the solution you'd like

Proposed behavior : in the minting form when minting gif files, have the option to choose a thumbnail. 
This would allow artists to set a static image or a reduced frame rate/quality gif.
Similar to how it currently works with video files

### Additional context

suggested by `@Cruel_Coppinger#0172` on discord: https://discordapp.com/channels/908672304236625970/910291653791735818/1090710369892323338
(https://twitter.com/cruel_coppinger)

melMass:
just to clarify: GIFs are already "thumbnailed" by the worker, but there is indeed a double upload of GIFs tokens. Fortunately the metas are still valid but the user upload twice for nothing

PR #350, opened 2023-10-27
fix: avoid uploading gif files twice

Zir0h:
There's no thumbnail for gif files, so avoid the second upload and reuse the same cid as the original.
Possible fix for https://github.com/teia-community/teia-ui/issues/278

//edit: thumbnails are managed by imgproxy anyway


melMass:
I'm dumb 🤦 ... But the actions take so long to run and I had all the PR open in new tabs

melMass (approved):
⭐

Issue #306, opened 2023-06-21
[bug] The filetypes described in the minting form fopr cover images are incorrect


merchantcoppola:
### Describe the bug

when minting files that allow cover images (i.e. music in this example)
the minting form says it allows all kinds of files as cover images
but actuaqlly only images are allowed:

```
export const ALLOWED_COVER_MIMETYPES = [
  MIMETYPE.JPEG,
  MIMETYPE.PNG,
  MIMETYPE.GIF,
  MIMETYPE.MP4,
]
```

<img width="825" alt="Captura_de_pantalla_2023-06-21_a_las_10 03 20" src="https://github.com/teia-community/teia-ui/assets/97635650/0128203b-d6bf-4819-8c8a-3c969607be7d">

### Reproduction

mint audio file

### Expected behavior

either adjust the explainer text to reflect the file tyopes that are allowed, or allow more filtypes

### Platform and versions

```shell
n/a
```

Zir0h:
Fixed with https://github.com/teia-community/teia-ui/pull/352

PR #352, opened 2023-10-27
fix: different validation rules for the cover file

Zir0h:
Fix for https://github.com/teia-community/teia-ui/issues/306


melMass:
Awesome!
I'm going to hold after reviewing jag's PR on that one since it's part of something I want to refactor there too, basically adding extra fields (data) instead of extra cases. The fields could be `allowed_types`, `error_label` etc .. The idea being providing extra information to "primitive" cases

melMass (approved):
Perfect in the meantime (regarding my previous comment) 😚

Issue #288, opened 2023-04-06
[bug] Minting preview does not work for Interactive OBJKTs


ktorn:
### Describe the bug

When minting an Interactive OBJKT, the preview fails to render the piece.

It seems that the OBJKT is actually rendered and runs (for example, the piece can log to the console), but it is placed in an iframe of 0x0 dimensions.

I found that manually overriding the iframe's container with `width=100%, height=100%` temporarily fixes the preview.

![CleanShot 2023-04-07 at 02 30 07](https://user-images.githubusercontent.com/119677/230465017-34328379-1250-40b5-b029-047e02326fe1.png)

### Reproduction

1. Go to MINT
2. Fill in all the details as usual
3. Upload an Interactive zip file (HTML/JS) OBJKT (for this test I used one of the templates from Teia)
4. Hit "Submit" to proceed to the Preview

### Expected behavior

Expected the Interactive OBJKT to render on the Preview screen

### Platform and versions

```shell
Mac OS 13.2.1
Chrome 111.0.5563.146
```

Zir0h:
should be fixed with https://github.com/teia-community/teia-ui/pull/354

PR #354, opened 2023-11-05
fix: re-enable preview for interactive objkts

Zir0h:
Fixes regression from https://github.com/teia-community/teia-ui/pull/346
Don't know if this is the right way of doing it though :sweat_smile:


Zir0h:
Also fix for https://github.com/teia-community/teia-ui/issues/288

Zir0h:
Maybe at some point [URL.revokeObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static) still needs to be called to free the used memory from `URL.createObjectURL` ?

melMass:
> Maybe at some point [URL.revokeObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static) still needs to be called to free the used memory from `URL.createObjectURL` ?

Yes I think too, otherwise it looks good, I would need to actually properly test it since all this is quite finicky

Zir0h:
I think this could be merged too? I mean, it's already broken now, so it would be challenging to break it even harder :laughing:

melMass:
True, and if it's breaking too much we can always revert it

melMass (approved):

Zir0h:
> True, and if it's breaking too much we can always revert it

what could possibly go wrong, right? :rofl:

Issue #220, opened 2022-12-03
[bug] PDF viewer flashing when paging


LeonNicholls:
### Describe the bug

The screen flashes between each page when you quickly click on the "next" link for a PDF OBJKT.

### Reproduction

1. Go to: https://teia.art/objkt/799229
2. Click on the "next" link for the PDF viewer
3. The screen flashes every time you click (especially when you click fast)

### Expected behavior

No flashing when paging a PDF.

### Platform and versions

```shell
Ubuntu 22.04 and Windows 10
Chrome Version 107.0.5304.121 and Brave Version 1.45.133
```

melMass:
Agreed very disturbing

v32k:
I was having a look and needed to untangle a few things to understand what the current state is.

First is that react-pdf doesn't seem to reuse canvases, or is just clearing and drawing in separate events I'm not really sure. This caused the primary flicker and seems very unlikely to change.

The workaround they give is to leave the previous page up until the next loads. The code currently used likely started here: https://codesandbox.io/p/sandbox/react-pdf-prevent-flash-with-scale-nse51l

This fixed one flicker but introduced another for the same reason of updates not happening simultaneously. Now there is page layout flicker from temporary DOM elements pushing things around.

A hack that can be done is to force all Page elements to be overlapping or carefully hiding overflow. The stylesheet in `components/media-types/pdf` looks suspiciously like someone tried this.

Setting `position: absolute` on both Page elements does work, but it makes pages zero size so nothing else flows around them. For controls to be placed correctly you have to manually set the height for the outer Document.

I might come back to this. I have an idea of what needs to happen I just don't know react at all.

PR #355, opened 2023-11-05
fix: prevent pdf flickering

Zir0h:
Possible fix for https://github.com/teia-community/teia-ui/issues/220
Looks like the page is rendered twice when doing next/previous, this causes it to jump around on the UI


melMass:
I can still see the flicker, what you removed was actually an attempt at removing the flicker (that failed).
I just tested it briefly, but with this PR all page flicker, whereas in master page flicker only once but once cached they don't anymore

melMass:
It's like 1 frame 😅 
pdf

Zir0h:
> I can still see the flicker, what you removed was actually an attempt at removing the flicker (that failed). I just tested it briefly, but with this PR all page flicker, whereas in master page flicker only once but once cached they don't anymore

aah okay! could it be related to the pdfjs version maybe? we're currently 2 major versions behind it seems :sweat_smile:

Zir0h:
In the [pdfjs examples](https://jsfiddle.net/pdfjs/wagvs9Lf/) they use a rendering queue, see `queueRenderPage`

melMass:
> we're currently 2 major versions behind it seems 😅

Yeah I think that would be wise to update 😅 

> In the [pdfjs examples](https://jsfiddle.net/pdfjs/wagvs9Lf/) they use a rendering queue, see queueRenderPage

This looks like the proper implementation of my attempt

Zir0h:
updating didn't change much, it still goes like page 1 => blank page for a flash => page 2 => blank page for a flash => page 3, it's weird

Zir0h:
The [example](https://projects.wojtekmaj.pl/react-pdf/) in react-pdf also flickers :laughing:

melMass:
> The [example](https://projects.wojtekmaj.pl/react-pdf/) in react-pdf also flickers 😆

Ohhh indeed ahah🤦 !
But the one you posted earlier is a good reference.
I will have time to dive and test more tomorrow

Zir0h:
> > The [example](https://projects.wojtekmaj.pl/react-pdf/) in react-pdf also flickers 😆
> 
> Ohhh indeed ahah🤦 ! But the one you posted earlier is a good reference. I will have time to dive and test more tomorrow

[This](https://github.com/wojtekmaj/react-pdf/issues/94#issuecomment-363197025) looks like what you were trying to do before? With some hidden page shenanigans :laughing:

Zir0h:
> > we're currently 2 major versions behind it seems 😅
> 
> Yeah I think that would be wise to update 😅
> 
> > In the [pdfjs examples](https://jsfiddle.net/pdfjs/wagvs9Lf/) they use a rendering queue, see queueRenderPage
> 
> This looks like the proper implementation of my attempt

This code also has the same issue. I think that the minted pdf files on teia are more "image" type than "text", so rendering them is always harder on the browser than just ASCII :/

Zir0h:
objkt.com uses https://www.npmjs.com/package/ng2-pdf-viewer which works very well, but I suppose there's no way to make that work on react?

//edit: without using iframes (because that would be easy I think, like we do for previews with iframe.teia.art)

Zir0h:
https://vadimdez.github.io/ng2-pdf-viewer/ this is snappy :smile: just put an ipfs link as source and uncheck "show all pages"

melMass:
> [vadimdez.github.io/ng2-pdf-viewer](https://vadimdez.github.io/ng2-pdf-viewer/) this is snappy 😄 just put an ipfs link as source and uncheck "show all pages"

Angular is widely different not sure how we could do that sanely tbh.

melMass:
Will merge tomorrow if no one see objections.

@Zir0h just migrated the iframe thing we removed a few months back to it's own repo -> https://github.com/teia-community/teia-iframe 

The idea proposed by Z0 would be to add support for pdf there so that we can use whatever work best

Issue #209, opened 2022-11-06
[feat] Display tezos domain instead of tezos address when available


LeonNicholls:
### Describe the problem

Currently, Teia displays the Tezos account address as part of the artist profile information. 
Objkt.com displays the Tezos domain  (*.tez) instead of the address when available.

### Describe the solution you'd like

Check if a Tezos domain is registered for an address and display that as part of the profile.
(Instead of the UI code doing the API calls, the indexer might do this instead)

### Additional context

Example:
https://teia.art/maxdrekker
https://objkt.com/profile/maxdrekker/created

PR #356, opened 2023-11-07
feat: show tezos domain on profile

Zir0h:
possible fix for https://github.com/teia-community/teia-ui/issues/209


src/pages/profile/profile.jsx
```diff
@@ -5,6 +5,27 @@ import Identicon from '@atoms/identicons'
 import styles from '@style'
 import { useDisplayStore } from '.'
 import ParticipantList from '@components/collab/manage/ParticipantList'
+import { useCallback, useEffect, useState } from 'react'
+
+async function reverseRecord(address) {
+  const response = await fetch('https://api.tezos.domains/graphql', {
```
melMass: I don't think this is standardized across the code base but we should use axios since I'm not sure we polyfill fetch yet (for older browsers)

Zir0h: switched to axios with https://github.com/teia-community/teia-ui/pull/356/commits/d04ad54692850530506fcb1f3fd31d72f33a9fac

melMass (approved):

PR #357, opened 2023-11-07
ci 🤖: pr preview url handling

melMass:
🤦 
- [ ] try to fix the router
- [ ] use the proper hash for the PR commit in the header (it seems to use a temporary one made for gh pages)

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -146,7 +146,10 @@ const router = createBrowserRouter(
         {display_routes}
       </Route>
     </Route>
-  )
+  ),
+  {
+    basename: import.meta.env.BASE_URL,
+  }
 )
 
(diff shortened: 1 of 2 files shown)


src/index.jsx
```diff
@@ -64,6 +64,11 @@ import { Preview } from '@components/preview/index'
 import MintForm from '@components/form/MintForm'
 import { ListsFeed } from '@pages/home/feeds/lists-feed'
 
+if (import.meta.env.BASE_URL !== '/') {
```
melMass: I'll have to double check but this should theoretically be automatically discarded in the builds of the real deploys

melMass (commented):
🚀🚀

PR #359, opened 2023-11-21
feat: update beacon

Zir0h:
the newer beacon version have the small wallet selection popup, this fixes the issue for people with smaller screens where they can't see the temple wallet option (it's hidden at the top)


melMass (approved):

melMass:
We might want to add "package.json" to the PR triggers at some point

PR #360, opened 2023-11-24
fix: run build command

Zir0h:
I was getting "Uncaught ReferenceError: require is not defined" applying this to vite.config seems to fix it

This is important because we do not use the builtin webserver to display the UI.

I found the solution on https://github.com/vitejs/vite/issues/3409#issuecomment-1138202247

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -137,6 +137,9 @@ export default defineConfig(({ mode }) => {
       port: 3000,
     },
     build: {
+      commonjsOptions: {
+        transformMixedEsModules: true
+      },
       outDir: 'build',
       sourcemap: !prod,
       rollupOptions: {


Zir0h:
I pushed an old image to prod, otherwise the users were greeted with a blank page :sweat_smile:

Zir0h:
Looks like this was fixed "the right way" in vite 5: https://github.com/vitejs/vite/commit/1061bbdaf81c2808ed8a86946cf0e07a0d8e5b13

melMass:
Why would we need to support `require`? I think the full codebase is ESM compliant

Zir0h:
> Why would we need to support `require`? I think the full codebase is ESM compliant

If you create a build with `npm run build` and then launch it with `cd build/; python3 -m http.server 9000` or with `docker compose up` you get a blank page. Could it be because of a dependency?

PR #367, opened 2023-12-16
Post-Mint Text: #SwapOnTeia

merchantcoppola:


melMass:
I just fear that might be noisy, maybe we could add a way to "don't show again"

Zir0h:
> I just fear that might be noisy, maybe we could add a way to "don't show again"

I'd rather have a bit too much noise in this case though :sweat_smile: it's like
 everybody just mints on Teia and then immediately swaps on objkt.

And the message itself is really not much:
2024-03-03-181951_1567x438_scrot

melMass:
> And the message itself is really not much

I can't check properly right now but doesn't this PR add an extra message box after mint? I would merge the "Mint done" and this one if that's the case, otherwise ignore my message

Zir0h:
> > And the message itself is really not much
> 
> I can't check properly right now but doesn't this PR add an extra message box after mint? I would merge the "Mint done" and this one if that's the case, otherwise ignore my message

It's the same box as before, just a different message

melMass:
That's what I couldn't check earlier, I'm still not sure since this is main currently: https://github.com/teia-community/teia-ui/blob/c7ae918563c58a6b19dc7772e3ce2e53f084189d/src/context/mintStore.ts#L306-L309

Zir0h:
> That's what I couldn't check earlier, I'm still not sure since this is main currently:
> 
> https://github.com/teia-community/teia-ui/blob/c7ae918563c58a6b19dc7772e3ce2e53f084189d/src/context/mintStore.ts#L306-L309

oh woops, didn't check that

Zir0h:
ah yeah, look at merch's first commit: https://github.com/teia-community/teia-ui/pull/367/commits/2bd6561a5122f1471af98d2c39ed7c3746e012ac

Zir0h:
I moved the message to prevent the double popup. There is however a little issue with my swap link. There's a small time frame between the tzkt indexer completion and the teia indexer completion. So the minter can get the "metadata being processed" if they click my swap link too soon. I don't think this is a big issue. It's possible to workaround this by also polling the teia indexer to check if the objkt is there, or maybe by adding a refresh on the "metadata being processed" page?

Zir0h:
replaced by https://github.com/teia-community/teia-ui/pull/502

Issue #373, opened 2024-01-10
[feat] Display OBJKTs that were flagged by the Teia content moderation team as NSFW or photosensitive.


LeonNicholls:
### Describe the problem

- Artists might not be aware that the Teia content moderation team flagged their OBJKTs as either NSFW or photosensitive.
- Artists don't always use the NSFW or photosensitive metadata flags when appropriate.

### Describe the solution you'd like

Add a feature to the UI to show any OBJKT's per account that the Teia content moderation team decided to flag as NSFW or photosensitive when missing from the minted metadata. 

It could provide transparency but also nudge the artist to use the flags. This information would be limited to only display to the account holder instead of all public users.

PR #378, opened 2024-02-02
feat: Display OBJKTs that were flagged by the Teia content moderation team as NSFW or photosensitive.

Zir0h:
Very ugly fix for #373


melMass:
This is d59703f98c0fa14e82a1b99859a62d7d4a12986c
image

melMass:
**only show it for yourself** in cbf47b8fff3f990999d85b3de6d46f034b80547d

**moderation corner** (added in c00872e407b1e1e26471b7195151f798040ed055)
image

melMass:
If we remove the BG style:
image

Zir0h:
Nice! Looks way cleaner without the background I think!
Thanks for the help!

Zir0h:
Maybe we also need to check when the objkt was minted, because the nsfw/photosensitive attributes didn't exist at the start

melMass:
What do you mean? I mean how is this PR affecting that? I think we normalize it all and old mints can still be moderated

Zir0h:
If I understood https://github.com/teia-community/teia-ui/issues/373 correctly, it's meant nudge repeat offenders of untagged nsfw/photo-sensitivity in the right direction.
It feels a bit unfair if you minted work really early and now you're seeing "you forgot to tag this, so we had to do it for you" all over your profile :sweat_smile:

melMass:
I see what you mean now!
Not sure we should complexify it just for that tbh, but I'm open if it's a real issue

Zir0h:
> I see what you mean now! Not sure we should complexify it just for that tbh, but I'm open if it's a real issue

Indeed, keep it simple :smile:

melMass:
I think now that's it's a simple corner with text it's ok, with the stripped red borders maybe not 🤣

melMass (approved):

PR #384, opened 2024-02-12
feat: check response body of rpc nodes

Zir0h:
otherwise a node that's unsynced will also be marked healthy because it responds with http/200


melMass:
I really need to dig more into nginx it's the last bits of code in this repo that I don't properly understand

melMass (approved):

Zir0h:
Best to merge this (and any other config changes that don't trigger the docker build) just before the "real" PR's. Otherwise the answer bot will complain that the commit hash isn't the latest on teia.art :smile: 
So I'm leaving this open for a bit, these changes have already been pushed to teia.art anyway.

melMass:
@Zir0h I let you decide when it's best to merge this one.
I think once #386 is done we can merge quite a few open PRs

Zir0h:
> @Zir0h I let you decide when it's best to merge this one. I think once #386 is done we can merge quite a few open PRs

yeah, this can be merged any time, but I'd do it before any of the other merges because adapting this config doesn't trigger a docker build and then answer bots yells because teia.art isn't running the latest commit :laughing:

PR #385, opened 2024-02-14
fix: load svg icons in firefox

Zir0h:
Before:
2024-02-14-223213_2513x934_scrot

After:
2024-02-14-223230_2507x958_scrot

I only see this issue in firefox though, chrome seems to handle the old css just fine.

I got the code from https://github.com/elad2412/the-new-css-reset/issues/66#issuecomment-1528855050


melMass:
I"m not sure to understand the comparison

Zir0h:
Only the first "tz" icon loads on firefox, I've tested this on windows and linux. I don't think it ever worked for me, but it never bothered me :smile:

melMass (approved):

PR #386, opened 2024-02-21
Anaverse 3D preview

Anaglyph-ic:
This PR adds a button next to the "fullscreen" button. 

Clicking the button replaces the asset with an iframe loading a 3D context which places the asset in its midst.

Clicking again replaces the iframe with the asset.


Zir0h:
Pretty cool addition, thank you!

We can also add this as a link right next to "View on ipfs", what do you think @melMass ? To iframe or not to iframe? :laughing:

Anaglyph-ic:
Changing the var to let, which is a apparently a JSX requirement.

melMass:
Thanks @Anaglyph-ic !
Yes this is just some linting rules.
I can have a proper look tomorrow morning!

Anaglyph-ic:
Happy to answer any questions if needed!

melMass:
I will review/refactor it now! This is very cool.
 
I would like to avoid dom manipulation like that so I'm just going to copy the logic in a component and conditionally display one or the other, this is a bit more "reacty".

Also toggling between the anaverse <-> teia views doesn't seem to work for 3D ones currently it should solve that.

Let's also add a tooltip, we don't currently have any specific tooltip I think so I'll just use the core html "title" for now:

melMass:
@Anaglyph-ic can you add me with write access to your repo? There is something strange and although I see you enabled maintainers edits, I still can't push to this PR

bd9dbe305b9dc9f98c8242cb107834210cc9edee

Anaglyph-ic:
Just added you! Thank you for your help!

Anaglyph-ic:
Thank you for the reactification!

melMass:
@Anaglyph-ic No worries, I think it works well, but in some instances not touching the mouse "drifts" the view to the right continuiously, I'm guessing it might be due to the fact that I used relative/`%` units to match our default viewers. Do you know if we can do something?

I'll add now @Zir0h 's suggestion, (an extra button to open `_blank` to anaverse).

Anaglyph-ic:
@melMass I personally think it's very tricky to not use absolute units (px) for such complex iframes. The anaverse works okay on smartphones, but maybe this option is too far-fetched for screens that are less than 800px wide? Your call, but maybe an option would be to keep it absolute-sized and deactivate it if `window.innerWidth < 800` ?

melMass (approved):
ready for me :)

merchantcoppola:
(for papertrail) here is the announcement blogpost: https://blog.teia.art/blog/anaverse-integration-announcement

Issue #387, opened 2024-02-23
[feat] / bug? swapped objkts dissapear from collection


Zir0h:
### Describe the problem

I just noticed that when you list a piece that you have collected to swap, it sort of disappears from your collections page.  I wasn't expecting that.  I'd kind of like to see it there in my collections page still unless someone else buys it.  I guess because it's put onto the Teia marketplace address then it doesn't appear on your collections page anymore?  I wonder if there is way you can see all the items you have collected and are reselling on a page?

-floydwilde on discord

### Describe the solution you'd like

Objects stay visible in collection

### Alternatives considered

Maybe a page with all swapped objects for your wallet?

floydwilde:
Might be better to wait until #383 is closed to look at this more.

Zir0h:
Fixed with https://github.com/teia-community/teia-ui/pull/388

PR #270, opened 2023-03-04
[current develop]

melMass:
> **Note**
> This is mostly ready, the biggest thing left is the filter bar that need some check / styling fix

deployed on: [preview.rocks](https://preview.teia.rocks)

- [x] 1/1 feed
- [ ] Bring back the profile feed filters (wip)
- [x] Add `tokens I own from this artist` filter (called `owned`, visible on the creation feed of an artist)
- [ ] It should already work, but check that it does now (#251)

PR #388, opened 2024-02-23
Also display swapped objkts in collection

Zir0h:
Maybe fix for https://github.com/teia-community/teia-ui/issues/387 ?


Zir0h:
Probably already fixed in develop? https://github.com/teia-community/teia-ui/pull/270/commits/11ffbbce389abb5f0e8b7ab146c1af4b7f0ff4d4

floydwilde:
I'm not sure there is any reason to be filtering on if an item is for sale or not on that page.  We only want to filter on items that the user owns that they have not created.  Whether it's for sale or not doesn't matter, I've proposed to remove those unneeded filters in: #389

melMass (approved):
LGTM

Issue #383, opened 2024-02-06
[feat] Add Filter functions on both Collection, Created


Zir0h:
### Describe the problem

Add Filter functions on both Collection, Created. Ie; filter by file type, date of mint,  recently listed, price, primary vs secondary listing, objkt vs teia contract, etc.
-Daniel Kantor via discord

### Describe the solution you'd like

n/a

floydwilde:
Was #426 meant to address this? I don't think it's resolved it if it was.  I just tried swapping a token I collected, and it's disappeared from my collections page when I did that: https://teia.art/objkt/829600

Zir0h:
> Was #426 meant to address this? I don't think it's resolved it if it was. I just tried swapping a token I collected, and it's disappeared from my collections page when I did that: https://teia.art/objkt/829600

That PR was bugged :sweat_smile: it kept showing tokens that were once owned by the account too, so still not  fixed I'm afraid.

PR #389, opened 2024-02-23
fix issue #387 collected items for sale visibility

floydwilde:
As mentioned in issue #387, @Zir0h pointed out where in the code this was likely happening. I read through the code for the collections page and could not understand the need to filter on FILTER_FOR_SALE and FILTER_NOT_FOR_SALE as it didn't align with the functionality we wanted for the page. The main goal here is to display all items a collector has acquired, irrespective of whether these items are currently listed for sale or not. This ensures a more consistent and user-friendly experience, where collectors can view their entire collection in one place.

In this pull request, I have removed the FILTER_FOR_SALE and FILTER_NOT_FOR_SALE filters and replaced them with a single FILTER_COLLECTED filter. This new filter effectively shows all tokens collected by the user that they didn't create themselves. The change simplifies the filtering logic and aligns the collections page with the intended purpose of showcasing all acquired items, making it easier for users to manage and view their collections.

This update should resolve the issue where tokens disappeared from the collection upon being listed for sale, as now all collected items remain visible regardless of their sale status.  I'm not that good at JavaScript though, so someone should review this for any syntax issue.

### DIFF
--- a/src/pages/profile/collections.jsx
+++ b/src/pages/profile/collections.jsx
@@ -8,8 +8,7 @@ import { BaseTokenFieldsFragment } from '@data/api'
 import { useOutletContext } from 'react-router'
 
 const FILTER_ALL = 'ALL'
-const FILTER_FOR_SALE = 'FOR_SALE'
-const FILTER_NOT_FOR_SALE = 'NOT_FOR_SALE'
+const FILTER_COLLECTED = 'FILTER_COLLECTED';
 
 export default function Collections() {
   const { showFilters, showRestricted, overrideProtections, user_address } =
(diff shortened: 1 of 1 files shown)


floydwilde:
dOh! I realize this is probably conflicting with development in #383 now, feel free to cancel or mark as draft and I could have another look after that work is completed.

melMass:
Thanks Floyd, yes those are not used for now because filtering is not fully implemented so we decided to hide it.
Before we migrated the infra it was allowing to display the sale status from any feed.
The new system's goal is to allow more fine tuned filtering with price range, dates, etc etc but it need quite some work, there is a branch IIRC that is about that where a few filters are working

PR #395, opened 2024-03-14
Update index.module.scss

Theohorsmeier:
added min-height to audio to fix the audio player having zero height


melMass:
Thanks!

I wonder if it could be a bit smaller though, here is 32px:

image

vs 54px:

image


It might be nice to apply theme too, but it can be a separate PR

Theohorsmeier:
Oh yeah 32 does look better tbh ..
I mostly picked 54 because that seemed to have been default somewhere 🫣

Issue #147, opened 2022-08-03
[feat] Support for MIDI (.mid) files on Teia


ryangtanaka-org:
### Describe the problem

MIDI is an underappreciated medium - it's lightweight, code compatible, and opens up a lot of opportunities for experimentation - perfect for TEIA to support its mime type, imo.

This was a long time ago but I did some work with this person for my grad school composition project - MIDI is interesting because it lets you visualize and manipulate the outcomes in ways that you can't with regular sound files. Might give some ideas about visualization as well. (He was doing this since the 80s, FYI.)

http://www.musanim.com/Player/

### Describe the solution you'd like

https://github.com/cifkao/html-midi-player

MIDI is a very archaic tech so a lot of it is unsupported but this one seems to be active -- might be fun to play around with the visualization aspects too 😎

melMass:
Hey Ryan

It's an interesting topic. 
The thing is that this is already doable using x-directory but like GLB it could indeed have basic support.
What would a "generic" midi player look like?

I personally like the idea that the artifact is the output file directly (MIDI, GLB) and not a web project export

ryangtanaka-org:
Hey, thanks for taking a look! Yeah, the html thing was just a suggestion (might be something we use for our own project) but if Teia wants to display the .mid files in a more neutral way I think that's totally cool too.

Most people probably don't know about the file format but for musicians (especially composers) it's an important tool for music-making that often doesn't get the attention it deserves - it'd be great if we could create a space for people to mint their works for posterity - I know a lot of folks in the classical music realm would probably appreciate it, if anything.

melMass:
Oh no, on our side it will be handled using some library like the one you provided. I was making the analogy to glb as this would work like them: stored in the native format on IPFS (we use the google viewer I linked you to on discord for that).

But I'm still unsure of the expectations of a basic MIDI player like is sound mandatory? Does it have to have a play function then or just display a table of notes?

ryangtanaka-org:
Sorry took a while to get back - MIDI is sort of a weird hybrid between computer code and sound so there was never really a "standard" way to "display" them, if you know what I mean. The interesting thing is that because it's just a series of commands, the way the music is *rendered* changes from machine to machine, depending on what type of soundcard or soundfont library they're using client-side. (Kind of fits in the ethos of Web3's decentralization mantra, but that's another issue altogether.)

[...]

PR #410, opened 2024-04-14
Feat: Add support for .mid file mints

kittypurrry:
This PR adds support for minting `.mid`, `.midi` file types in the `MintForm` component.

**Preview / Token Page**
_Example: `https://localhost:3000/objkt/853348`_

The preview component displays a Midi Player, and the auto ggenerated cover image, inline with Teia's current UI.

image

**Feed Page**

Midi NFTs are displayed similarly to Audio NFTs on Teia's feed listing page. Clicking on the play button plays the .mid soundfile.

image

**Notes**
- There is an issue where mid files are not automatically stopped when navigating away from the page.  The workaround was to add a disclaimer text to get users to manually stop the player if they are navigating away from the page while the .mid file is playing. 
- It seems like OBJKT does not have support for playing .mid files, something that we could possibly add to the OBJKT feature request? See: https://objkt.com/tokens/hicetnunc/850481 on a .mid token. 

--

_Update: 21/04:_
- Used HTML Midi Player (https://cifkao.github.io/html-midi-player/) instead of JZZ to allow us to control velocity. 
- Added volume slider next to midi player that modifies the 'velocity' of the midi file
- Note: Since we have declared a `global` definition and set it as `globalThis` in the `vite.config.js`, this caused some issues when building because some of the files in `@magenta/music` as required by the HTML midi player package is importing a file with the name `global` (i.e global.js). The rules in `vite.config.jq` automatically resolves such file names to 'globalThis' and this causes an error since the file cant be found. Added aliases to the `vite.config.js` in order to workaround this issue. 

_Update 16/06:_
- Create function `generateMidiCover` to automatically generate midi covers based on the MidiJSON object tracks and notes within each track. Each track will be represented in a different opacity shade of white based on the track # in the midi JSON object. Any mints with mime type `audio/mid` or `audio/midi` will no longer be required to upload a cover image file.

image


melMass:
Thanks for the contribution!!

This would close #147 cc @ryangtanaka-org

If no one did I can review this tuesday or the next day

Zir0h:
I installed this on <https://preview.teia.art/objkt/850481>

melMass:
> This PR adds support for minting .mid, .midi file types in the MintForm component.

It looks like something happened in your recent commits as I don't see the mint related edits. Only the token viewer for `.midi`

PR #428, opened 2024-10-15
Copyrightform debug

ryangtanaka-org:
- fixed date input bug where the date would reset itself to the default during input
- fixed UI bug where when non-transferable rights were unchecked, it would show up as undefined rather than false


Zir0h:
Deployed to https://preview.teia.art/

Zir0h (approved):

PR #442, opened 2024-12-12
fix: ⚡Add teia.rocks wildcard, teztok, and connkt URLs for Interactive OBJKTs

ktorn:
Adding URLs used by networked / interactive OBJKTs.

### DIFF
--- a/src/utils/html.js
+++ b/src/utils/html.js
@@ -195,7 +195,7 @@ export function injectCSPMetaTagIntoHTML(html) {
       https://*.tzkt.io
       https://api.hicdex.com
       https://hdapi.teztools.io
-      https://api.teia.rocks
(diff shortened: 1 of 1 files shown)


Zir0h:
Thanks for the PR!

I think it would be better to just get rid of injectCSPMetaTagIntoHTML altogether.

For example hicetnunc.art should actually be treated as "unsafe" in my opinion, that domain name expired and got scooped up by someone else :/ this list can also modified by the uploader anyway.

The correct place to handle this security is on the webserver itself imho.

What do you think @xat @melMass ?

Zir0h (approved):

melMass:
Agreed, imo "security" things that aren't handled by the SC should be handled by the server and not the client, "forging" objkt is really not complicated

PR #449, opened 2025-02-02
Teia UI update

ryangtanaka-org:
- added donation button to the treasury, a quipuswap link to convert xtz into TEIA

- fixed incorrect contract for the treasury balance display - please confirm that (KT1RRLSxpsiKpnVZQsBGCqG9HFBXbPBtUYBD) is correct

- added core parts of the documentations into the top right hand corner menu items (the text is in lang/en if edits are needed)

- minor fixes in the feeds dropdown menu on the top left

- added subtitles to the swap tab so it's easier to see

- added explanations to the DAO and polling pages for more clarity

- FAQ page updates and added explanation for the copyright feature

- various UI fixes throughout the site


Zir0h:
Why did you add the markdown-to-jsx and fetch stuff in for example https://github.com/teia-community/teia-ui/blob/4df6007632a298632b63016bf00b90cdabdd0266/src/pages/codeofconduct/index.jsx ? The markdown rendering should also just render if you imported it the same way as for example https://github.com/teia-community/teia-ui/blob/main/src/pages/about/index.jsx

Zir0h:
I think treasury balance should be taken from KT1J9FYz29RBQi1oGLw8uXyACrzXzV1dHuvb (core team multisig) instead.

// and KT1J9FYz29RBQi1oGLw8uXyACrzXzV1dHuvb should also be the donation target address I think

src/components/header/feed_toolbar/FeedToolbar.jsx
```diff
@@ -62,9 +62,12 @@ export const FeedToolbar = ({ feeds_menu = false }) => {
   )
   const location = useLocation()
   const feedLabel =
-    locationMap.get(location.pathname) || startFeed || DEFAULT_START_FEED
+    [...locationMap.entries()].find(([key]) =>
+      location.pathname.startsWith(key.replace('*', ''))
+    )?.[1] || 'Sort Feed' // Default to "Sort" if no match is found
```
Zir0h: What is 'Sort Feed'?

src/components/header/feed_toolbar/FeedToolbar.jsx
```diff
@@ -98,7 +108,14 @@ export const FeedToolbar = ({ feeds_menu = false }) => {
...
                   return (
-                    <Button key={k} to={k}>
+                    <Button
+                      key={k.replace('*', '')}
+                      to={
+                        k.startsWith('/feed/friends*')
+                          ? `/feed/friends/${walletAddress}`
+                          : k.replace('*', '')
+                      }
+                    >
```
Zir0h: I don't understand this part :/
walletAddress is only filled in for logged in users

src/components/header/feed_toolbar/FeedToolbar.jsx
Zir0h: Answered in discord: this is to update the label wording, but it breaks the functionality a bit (it's always 'Sort Feed' when I test it). Maybe it's better to just prefix the selector with a "Sort by: " text or something?

src/components/header/feed_toolbar/FeedToolbar.jsx
Zir0h: Fixed with dirty workaround

Zir0h (approved):

PR #450, opened 2025-03-08
MintStore Input Fix (?)

ryangtanaka-org:
- Replaced the default values of the Form with mintStore, rather than the defaultValues, which was overwriting the user's input in certain cases

- Removed console logs and key errors from the CustomCopyrightForm

I think it's working but needs to be tested for stability.


Zir0h (approved):

Zir0h:
do you have some scenario's to test? would be good to see a before/after

ryangtanaka-org:
The scenario I encountered was when I would check the preview page before mint, realized needed edits, but when hitting the back button erased all the changes I made prior. It's been a while so don't remember a lot of it but I *think* it's fixed now?

PR #454, opened 2025-05-25
[Docs] FAQ update for curation tab in user profile

floydwilde:
Just mentioning that reswaps end up on the curation page.


floydwilde:
Oops I think I was on the wrong branch when I create this.

PR #460, opened 2025-08-25
Fix tokens search on copyright creation

josim:
Update swr/fetchTokenMetadata - renamed to fetchTokenMetadataForCopyrightSearch - use teia rocks api for hen tokens, tzkt for everything else Update CustomCopyrightForm & CopyrightDisplay to use new function Name


Zir0h:
pushed to https://preview.teia.art/

PR #478, opened 2025-11-18
Fix Issue #473 & Clean up NPM

LeonNicholls:
I have fixed the `npm test` command, updated project dependencies, eliminated all `npm install` warnings and security vulnerabilities, resolved runtime errors, and eliminated Vite startup warnings.

### 1. Fixed `npm test`
- **Dependencies:** Updated `vite`, `vitest`, `@vitejs/plugin-react`, and `jsdom` to compatible versions.
- **Configuration:**
    - Modified [package.json](file:///home/leon/Downloads/teia-ui/package.json) to use `vitest` for the test script.
    - Updated [vite.config.js](file:///home/leon/Downloads/teia-ui/vite.config.js) to include Vitest configuration.
    - Excluded Playwright tests from Vitest execution.
- **Verification:** Added [src/App.test.jsx](file:///home/leon/Downloads/teia-ui/src/App.test.jsx) to verify the test runner.

### 2. Eliminated NPM Warnings
- **Peer Dependencies:**
    - Updated Storybook packages to `^8.4.7`.
    - Removed unused `vite-plugin-node-stdlib-browser`.
    - Updated `vite-plugin-node-polyfills` and other plugins.
- **Deprecated Packages:**
    - Updated `canvas` to `^3.0.0` (via `devDependencies` and `overrides`) to remove `node-pre-gyp` deprecation warnings (`npmlog`, `are-we-there-yet`, `gauge`).
- **Deprecated Config:** Removed `unsafe-perm=true` from [.npmrc](file:///home/leon/Downloads/teia-ui/.npmrc).

### 3. Security Updates
- **Vulnerabilities Resolved:** Reduced vulnerabilities from ~78 to **0**.
- **Major Updates:**
    - Updated `react-pdf` to `^9.0.0` (breaking change).
    - Updated `playwright` to `^1.41.2`.
    - Updated `@taquito` packages to `^23.0.3`.
    - Updated `axios` to `^1.7.9`.
- **Overrides:**
    - Added overrides for `fast-redact`, `esbuild`, `pino`, `canvas`, `minimist`, and `static-eval` to force secure/modern versions.

### 4. Fixed Runtime Errors
- **`react-pdf` v9 Migration:**
[...]

PR #480, opened 2025-11-23
Fix packages warnings - fix beacon

kylegrover:
subfix for #478 fixes beacon, pins node to 24 with .nvmrc, and adds cross-env to package to fix windows build


Zir0h:
partly handled with https://github.com/teia-community/teia-ui/pull/506
we can update node 22 => 24 later

PR #484, opened 2026-02-07
Fix update beacon taquito

Zir0h:
here we go again


package.json
```diff
@@ -37,7 +37,8 @@
     "react-use": "^17.4.0",
     "react-use-clipboard": "^1.0.9",
     "screenfull": "^6.0.2",
-    "swr": "^1.3.0"
+    "swr": "^1.3.0",
+    "three": "^0.182.0"
```
kylegrover: was this just added or it's just a change of how it's included?

src/context/userStore.ts
```diff
@@ -595,17 +576,17 @@ export const useUserStore = create<UserState>()(
           const contract = await Tezos.wallet.at(
             proxyAddress || MARKETPLACE_CONTRACT_V1
           )
-          const mint_batch = contract.methods.mint_OBJKT(
-            tz,
-            amount,
-            `ipfs://${cid}`
+          const mint_batch = contract.methodsObject.mint_OBJKT({
```
kylegrover, line 579: have you tried a test mint?

src/context/userStore.ts
```diff
@@ -210,18 +206,15 @@ export const useUserStore = create<UserState>()(
           // This piece of code should be called on startup to "load" the current address from the user
           // If the activeAccount is present, no "permission request" is required again, unless the user "disconnects" first.
 
-          await wallet.client.requestPermissions({ network })
+          await wallet.client.requestPermissions()
```
kylegrover, line 209: do we need the { network } to get the custom rpc settings?

src/context/userStore.ts
Zir0h: yes, but more testing is always welcome

package.json
Zir0h: woops, this is probably me messing around

src/context/userStore.ts
Zir0h: good question, don't know the answer, and teia mainnet == teia testnet anyway (for now) :sweat_smile:

kylegrover: i think we should try adding it back in and if that works just keep it in, seems linked to the rpc settings and those are nice to have when theres network weirdness

Zir0h: it got the red underlining of an error in my editor though

Zir0h: https://github.com/airgap-it/beacon-sdk/blob/bf189c919fb226efed39c80edc23a060a67e8c74/packages/beacon-dapp/src/dapp-client/DAppClient.ts#L1517-L1524

PR #486, opened 2026-02-18
Teia blog feature

josim:
Summary                                                                                                                                                                                                            
                                                                                                                                                                                                                     Adds the community blog section (/blog) built on top of Teia's existing text/markdown minting infrastructure.
Blog posts are standard Teia OBJKTs.

What's new

  - Blog section (/blog) with four tabs:
    - Community, feed of all text/markdown and text/plain OBJKTs, newest first
    - Bulletin, official Teia posts, filtered by DAO multisig signer address + teiamultisigblog tag
    - Your Posts, authenticated user's own blog posts with edition counts and burn management (warns if active listings exist before burning)
    - New Post, full markdown editor (@uiw/react-md-editor, lazy-loaded) with live preview, draft auto-save to localStorage, cover image generation, tags, and edition/royalty configuration
  - Token embedding, custom editor toolbar command lets authors search for any OBJKT by ID, URL, name, or artist address and embed it in a post. Embedded token metadata (artist, royalties) is stored in a
  teia_embedded_tokens field on IPFS for downstream use.
  - Profile blog tab, new "Blog" tab on user profiles showing that user's blog posts, with burn controls on your own profile
  - Improved markdown display, the text/markdown media renderer now shows full rendered content inline instead of a cover image with a link. Also adds audio playback support within markdown (renders <audio> for
  audio file references).
  - "Blog" in main nav, new header link

  Technical details

  - Blog posts are queried via new SWR hooks (useBlogPosts, useBlogPostsByArtist, useOfficialBlogPosts) that filter the existing GraphQL token index by mime type
  - Official post detection uses useMultisigAddresses() which fetches DAO treasury contract signers from TzKT
  - Token search for embedding uses a new searchTokensForEmbed() GraphQL query that handles token IDs, teia.art URLs, tz addresses, and free-text name search
  - Embedded tokens are stored as HTML comments (<!-- teia-token:{...} -->) in the markdown source
  - Drafts are debounce-saved to localStorage (title, body, tags, cover)
  - @uiw/react-md-editor (^4.0.11) is the new dependency, lazy-loaded via React.lazy


kylegrover:
draft debounced localstorage 👍 

lazy load md editor 👍 tho can we make sure people are also able to edit and post in raw text if preferred? I'm also thinking some artists might prefer to control which of their md/text posts are displayed as 'blogs' or not.
 
Official post detection - I think this should be adjusted to query for posts that were made by current and past multisig members while they were members (and not show blog posts they made before or after they were in the multisig). All the data should be on chain but I'm imagining some index updates might be needed to pull that off.

Markdown objkt/page/links - this is an interesting idea, can you talk more about the design? what should happen when a token or multiple tokens are embedded, or an artist or url? should they be displayed fully inline? linked? do we want to handle the display in some special way? are there any use cases / user stories in mind for the feature?

josim:
@kylegrover thanks for the feedback!

**lazy load md editor:** Yeah right now every text/markdown or text/plain token shows up in the community feed. While there is still the blog tab on the profile page, anyway, we could add a filter for token types.

**Official post detection:** There might be a way to query this via tzkt, I will check on it. Keeping this info in a json file might be good enough since it's not a lot of data (with timestamp).

**Markdown objkt/page/links:** 
Once a Teia token has been added to the Markdown editor, it is displayed in the preview with the displayUri/cover for videos/audio (audio also displays a player) / 3D objects  / text / blogs (any type should be covered) and the walletaddress/alias.
Additional, the wallet address of the artist of the linked token is stored in a special field in the token metadata: “teiaEmbeddedTokens” with a royalty share of 15%. 
The idea behind this is that the associated artists receive a share when a blog is sold.
Here is a test blog: https://preview.teia.art/objkt/877405

Zir0h (approved):

PR #512, opened 2026-05-18
Add verified Bluesky links to profiles

tulpenhaendler:
## Add verified Bluesky links to profiles

> 🚧 **WIP — keeping this as a draft for now.** Kindly requesting a review of the feature on this branch; I'm open to suggestions on any of it (UX, the verification approach, where the code lives, etc.).

Tezos Commons is about to launch [tzbsky.com](https://tzbsky.com) as part of an effort to bring the Tezos community onto ATProto / Bluesky. The service lets people link their Bluesky and Tezos/Etherlink accounts, which opens up things like custom badges, Tezos-only feeds, and other community goodies on the Bluesky side.

It also lets a user **publicly** link a Tezos address to their Bluesky account (their ATProto DID) by publishing a signed record to their own Personal Data Server (PDS). tzbsky.com additionally exposes a reverse-lookup API over those public records.

This PR adds a small Bluesky icon to any profile whose owner has published such a link — so visitors can jump straight to their Bluesky account. **The link is verified client-side before we show it**: we don't just trust the lookup API, we check the on-PDS signature ourselves.

A working example, when running this branch locally: http://localhost:3000/tz/tz1cxfPsZ2h74snLzh6MkjhmezvekW9vog5B

### Quick ATProto primer: DID and PDS

Bluesky is built on ATProto, where two concepts matter here:

- **DID** (Decentralized Identifier) — the *stable* identity of an account, e.g. `did:plc:abc123…`. It never changes, even if the user renames their handle or moves servers. The human-readable handle (`@alice.bsky.social`) is just a mutable pointer *to* a DID, which is why we always link and verify against the DID and treat the handle as a display label only. A DID resolves to a small JSON document that says, among other things, which server currently hosts the account's data.
- **PDS** (Personal Data Server) — the server that actually stores a user's repository: their posts, profile, and any other records, including the Tezos↔Bluesky link record. Crucially the PDS is *the user's own* (self-hostable, or Bluesky-hosted), and records there are signed by the account's repo key — so a record fetched from someone's PDS is something *they* put there, not something tzbsky or anyone else asserts on their behalf, tzbsky merely has a nice interface for doing this.

So the chain is: address → DID (who) → PDS (where their data lives) → the signed link record (the proof).

### How it works

For each profile we take the wallet address and:

1. **Discover a candidate DID** via the tzbsky reverse-lookup API. This is treated as an *untrusted hint* — a bad answer here can't fool us, it just makes the next steps fail.
2. **Resolve the DID to its PDS** (`plc.directory` for `did:plc`, the domain's `did.json` for `did:web`) and **fetch the signed record** (`com.tzbsky.cryptoAddress`) straight from the user's own PDS.
3. **Verify it ourselves**, two independent checks, both required:
   - **Binding** — the signed message must name *this* DID, *this* address, and `chain=tezos`. This is what stops a valid signature being lifted from somewhere else and replayed into another repo.
   - **Signature** — verified as described below.
4. **Resolve the current handle** from the DID, purely for display — the handle is mutable and never part of the proof, so the link points at the stable DID and the handle is just the label.

If anything fails — no record, bad signature, network hiccup — we simply don't show the icon. Nothing is cached: there's no cryptographic revocation, so the record is re-fetched and re-verified on every profile view. The badge links to `bsky.app/profile/<did>`.

### The signature

The record stores the wallet signature over a plain-text attestation message (the lines naming the DID, address, chain, etc.). To verify it we have to reconstruct the *exact bytes the wallet signed*, which is a Tezos off-chain signed message — i.e. a **Michelson value packed in the PACK format**:

```
0x05 0x01 <4-byte big-endian length> <utf-8 message bytes>
```

`0x05` is the PACK magic byte, `0x01` is the Micheline `string` tag, then the length-prefixed UTF-8 message. We build that byte string, then hand it to `@taquito/utils` `verifySignature`, which Blake2b-256 hashes it and checks the signature against the public key on the appropriate curve (Ed25519 / secp256k1 / P-256, picked from the key prefix). Separately we derive the address from that same public key (`getPkhfromPk`) and require it to equal the address we're looking up — so the key that signed is provably the key that owns the wallet, not just *some* valid key over the message.

### On tzbsky being optional

Worth being explicit: **the proof does not depend on tzbsky.com at all.** The trust comes entirely from the signed record living in the user's own PDS plus the signature check above. If you maintained your own index of these public ATProto records (they're public by design), you could do this lookup with zero involvement from tzbsky — resolve the DID, fetch the record, verify. The reverse-lookup API is purely a convenience so we don't have to crawl and index every PDS ourselves; it makes the address→DID step easy, but it's not part of the security model and a wrong/missing answer there only causes the check to fail closed, never to falsely succeed.

### DIFF
--- a/.env
+++ b/.env
@@ -13,6 +13,11 @@ VITE_TEIA_DONATION_API='https://teia-donation-express-app-production.up.railway.
 
 VITE_TZPROFILES_GRAPHQL_API='https://indexer.tzprofiles.com/v1/graphql'
 
+# tzbsky: wallet address -> Bluesky (atproto) identity
+VITE_TZBSKY_API='https://tzbsky.com'
+VITE_PLC_DIRECTORY='https://plc.directory'
+VITE_BSKY_PUBLIC_API='https://public.api.bsky.app'
+
 VITE_IMGPROXY='https://imgproxy.teia.rocks'
 
 VITE_IPFS_UPLOAD_PROXY='https://ipfsproxy.teia.rocks'
--- a/src/components/header/feed_toolbar/FeedToolbar.jsx
+++ b/src/components/header/feed_toolbar/FeedToolbar.jsx
@@ -1,4 +1,5 @@
 // ! NOTE - Keep the comments.
+import { Fragment } from 'react'
 import { motion } from 'framer-motion'
 import styles from '@style'
 import { DropDown, DropdownButton } from '@atoms/dropdown'
@@ -88,12 +89,12 @@ export const FeedToolbar = ({ feeds_menu = false }) => {
                 {[...locationMap.keys()].map((k) => {
                   if (k.startsWith('-')) {
                     return (
-                      <>
-                        <Line className={styles.separator} key={k} />
+                      <Fragment key={k}>
+                        <Line className={styles.separator} />
                         <span className={styles.subtitle}>
                           {locationMap.get(k)}
                         </span>
-                      </>
+                      </Fragment>
                     )
                   }
                   if (locationNeedSync.includes(k)) {
--- a/src/pages/profile/profile.jsx
+++ b/src/pages/profile/profile.jsx
@@ -8,6 +8,7 @@ import { useDisplayStore } from '.'
 import ParticipantList from '@components/collab/manage/ParticipantList'
 import { useCallback, useEffect, useState } from 'react'
 import axios from 'axios'
+import { resolveVerifiedBluesky } from '@utils/bsky'
 
 async function reverseRecord(address) {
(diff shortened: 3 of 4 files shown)


tulpenhaendler:
the tzbsky.com UI for a bit more context, as soon as the user links the address in this specific setting, the Bsky icon would appear on the teia profile: 
image

tulpenhaendler:
dont think the pds recod format will change, good to go

Zir0h (approved):

Zir0h:
Thank you!

